PR #28, opened 2016-06-29
Refactor cljs api

spalladino:
Do not use core async channels for managing API response in projects creation.

(PR built on top of https://github.com/instedd/planwise/pull/27)

### DIFF
--- a/src/planwise/client/api.cljs
+++ b/src/planwise/client/api.cljs
@@ -19,7 +19,8 @@
 (swap! default-interceptors (partial cons csrf-token-interceptor))
 
 
-;; Common json-request definition to use with ajax requests
+;; Common request definitions to use with ajax requests
+
 (defn common-error-fn [{:keys [status status-text]}]
   (c/error (str "Error " status " performing AJAX request: " status-text)))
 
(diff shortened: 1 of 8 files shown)


src/planwise/client/api.cljs
```diff
@@ -19,6 +19,16 @@
...
+;; Default json request settings
+
+(defn error-handler [{:keys [status status-text]}]
+  (.error js/console (str "Error " status " performing AJAX request: " status-text)))
+
+(def json-opts
+  {:format :json
+   :response-format :json
+   :keywords? true
+   :error-handler error-handler})
```
spalladino: We could turn `json-opts` into a function that actually receives the params, success-fn, and any other settings, and merges (and optionally preprocesses them). It could be useful to add a new `mapping-fn` that wraps the `success-fn`.

src/planwise/client/projects/handlers.cljs
```diff
@@ -25,8 +25,9 @@
  :projects/create-project
  in-projects
  (fn [db [_ project-data]]
-   (async-handle (api/create-project project-data)
-                 #(dispatch [:projects/project-created %]))
+   (api/create-project
+     project-data
+     #(dispatch [:projects/project-created %]))
```
spalladino: If we turn `json-opts` to a function, we could allow passing a `:keyword` as `success-fn`, which is understood as dispatching that event using the response as data. So the above block would be:
```clojure
(api/create-project project-data :projects/project-created)
```

PR #31, opened 2016-07-12
Load projects when navigating to project URL

spalladino:


src/planwise/client/db.cljs
```diff
@@ -28,7 +28,10 @@
 
    ;; Projects
    :projects
-   {:creating? false}
+   {:creating? false
+    :creating-waiting? false
+    :loading? false
+    :current nil}
```
spalladino: Cache of `id => project` was replaced by a single `current` field holding the project being shown

src/planwise/client/db.cljs
```diff
@@ -28,7 +28,10 @@
 
    ;; Projects
    :projects
-   {:creating? false}
+   {:creating? false
+    :creating-waiting? false
+    :loading? false
```
spalladino: Loading is used when loading a project given its id

src/planwise/client/db.cljs
```diff
@@ -28,7 +28,10 @@
 
    ;; Projects
    :projects
-   {:creating? false}
+   {:creating? false
+    :creating-waiting? false
```
spalladino: `creating?` and `creating-waiting?` are actually representing a `creating-state` with 3 possible values. How should this be handled? Just symbols, or is there something closer to enums that can be used for this?

src/planwise/client/handlers.cljs
```diff
@@ -13,9 +13,24 @@
    (playground/fetch-facilities-with-isochrones :immediate (:playground db/initial-db))
    db/initial-db))
 
+(defmulti navigate-handler (fn [db [_ page & _]] page))
```
spalladino: Is a multimethod a correct choice here, rather than simply a large `case`?

src/planwise/client/handlers.cljs
```diff
@@ -13,9 +13,24 @@
...
+
+(defmethod navigate-handler :projects [db [_ page & params]]
+  (let [id (js/parseInt (first params))
+        db (assoc db
+             :current-page page
+             :page-params params)]
+    (if (not= id (get-in db [:projects :current :id]))
+      (do
+        (dispatch [:projects/load-project id])
```
spalladino: Loading the project is _requested_ by firing an event, which is handled appropriately. Should this be handled by actually setting the appropriate db fields and calling the API here, rather than delegating that to the handler?

src/planwise/client/projects/api.cljs
```diff
@@ -3,6 +3,16 @@
...
 
+(defn load-project [id]
+  (let [c (chan)]
+    (GET
+      (str "/projects/" id "/details")
+      (assoc (async-handlers c identity)
+            :format :json
+            :response-format :json
+            :keywords? true))
+    c))
```
spalladino: TODO: Refactor the API to remove channels boilerplate

src/planwise/client/projects/views.cljs
```diff
@@ -161,10 +165,20 @@
      [:h1 "Scenarios"]]))
 
 (defn project-view []
-  (let [page-params (subscribe [:page-params])]
+  (let [page-params (subscribe [:page-params])
+        current-project (subscribe [:projects/current])
+        project-goal (reaction (:goal @current-project))]
```
spalladino: Ok to use a reaction on a subscription within a component?

src/planwise/endpoint/projects.clj
```diff
@@ -11,5 +11,10 @@
       (let [projects (projects/list-projects service)]
         (response projects)))
 
+    (GET "/:id/details" [id]
+      (if-let [project (projects/get-project service (Integer/parseInt id))]
+        (response project)
+        (throw "not-found")))
```
spalladino: How should we appropriately handle 404s?

src/planwise/endpoint/projects.clj
```diff
@@ -11,5 +11,10 @@
       (let [projects (projects/list-projects service)]
         (response projects)))
 
+    (GET "/:id/details" [id]
```
spalladino: `/details` is required to avoid clashing with `/project/:id`, which is captured by the home endpoint. Should we move all json requests under an `/api/` context? Or route based on content-type? Or add a `.json` extension to all these endpoints?

PR #38, opened 2016-07-18
Filter facilities

fgasperij:


resources/planwise/sql/facilities.sql
```diff
@@ -12,6 +12,13 @@ SELECT
 FROM facilities
 ORDER BY name;
 
+-- :name facilities-with-types :? :*
```
spalladino: Should be `facilities-from-types`, not `with-types`. 
Also, shouldn't we also `SELECT` the `type` field in all queries?

src/planwise/client/db.cljs
```diff
@@ -39,8 +39,10 @@
                  :filters {:type #{}
                            :ownership #{}
                            :services #{}}
-                 :count 16
-                 :total 125}}
+                 :count 0
+                 :total 4944
+                 :facilities []} ;; TODO the path [:facilities :facilities] is confusing
```
spalladino: On projects we have `[:projects :list]` to avoid this problem, but a better name is also welcome

src/planwise/client/projects/handlers.cljs
```diff
@@ -81,5 +81,32 @@
...
+         updated-filters (into {} (for [[k v] updated-db-filters] [k (apply vector v)]))]
+     (api/fetch-facilities updated-filters :projects/facilities-loaded)
+     new-db)))
+
+(register-handler
+ :projects/facilities-loaded
+ in-current-project
+ (fn [db [_ response]]
+   (let [count-path [:facilities :count]
+         facilities-path [:facilities :facilities]
```
spalladino: Though it's a style choice, I'd avoid creating a binding for each of the paths, as they are quite simple and used only once

src/planwise/endpoint/facilities.clj
```diff
@@ -7,8 +7,10 @@
 
 (defn- endpoint-routes [service]
   (routes
-   (GET "/" [] (let [facilities (facilities/list-facilities service)]
-                 (response facilities)))
+   (GET "/" [type :as req] (let [facilities (facilities/list-facilities-with-types service (vals type))]
+                             (println req)
+                                      (println type)
```
spalladino: Remove the debugging statements!

src/planwise/client/projects/handlers.cljs
```diff
@@ -81,5 +81,32 @@
          current-filter (get-in db path)
          toggled-filter (if (contains? current-filter filter-value)
                           (disj current-filter filter-value)
-                          (conj current-filter filter-value))]
-     (assoc-in db path toggled-filter))))
+                          (conj current-filter filter-value))
+         new-db (assoc-in db path toggled-filter)
+         updated-db-filters (get-in new-db [:facilities :filters])
+         updated-filters (into {} (for [[k v] updated-db-filters] [k (apply vector v)]))]
```
spalladino: Move `vector` of sets inside `api/fetch-facilities`

PR #52, opened 2016-07-21
Feature/process after import

ggiraldez:
- Make the importer worker preprocess isochrones after importing the collection from Resourcemap
- Report import/process progress in the UI

### DIFF
--- a/resources/migrations/013-process_single_facility.up.sql
+++ b/resources/migrations/013-process_single_facility.up.sql
@@ -14,7 +14,6 @@ begin
     agg_cost double precision
   );
 
-  -- Process all facilities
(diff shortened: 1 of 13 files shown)


resources/migrations/013-process_single_facility.up.sql
```diff
@@ -0,0 +1,73 @@
...
+  from_cost integer;
+  to_cost integer;
+  facility_node integer;
+begin
+  create temporary table if not exists edges_agg_cost (
+    gid integer not null,
+    agg_cost double precision
+  );
+
+  -- Process all facilities
```
spalladino: Stale comment :-(

PR #76, opened 2016-07-25
Generic client application config mechanism

### LINKED ISSUE #67: Use Resourcemap URL from the config
In the datasets screen there's a link to Resourcemap currently hard-coded. That URL should be calculated from the configuration.

Also, open Resourcemap in a new window.

ggiraldez:
- Add a generic way for the server to communicate global/per request
  configuration to the client application.
- Use the config mechanism to retrieve the JWE token, user email and
  Resourcemap URL.

Closes #67.

### DIFF
--- a/src/planwise/endpoint/home.clj
+++ b/src/planwise/endpoint/home.clj
@@ -3,6 +3,7 @@
             [ring.util.anti-forgery :refer [anti-forgery-field]]
             [buddy.auth :refer [authenticated? throw-unauthorized]]
             [planwise.component.auth :refer [create-jwe-token]]
+            [cheshire.core :as json]
             [hiccup.form :refer [hidden-field]]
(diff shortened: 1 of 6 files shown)


src/planwise/endpoint/home.clj
```diff
@@ -3,6 +3,7 @@
             [ring.util.anti-forgery :refer [anti-forgery-field]]
             [buddy.auth :refer [authenticated? throw-unauthorized]]
             [planwise.component.auth :refer [create-jwe-token]]
+            [cheshire.core :as json]
```
spalladino: Why not `clojure.data.json/write-str`?

ggiraldez: Mostly for consistency. `wrap-json-params` and `wrap-json-response` use Cheshire already. Also it's faster: https://github.com/dakrone/cheshire#why :)

PR #107, opened 2016-08-01
Popups for markers

spalladino:
Change marker-layer to accept a collection of hashmaps along with fns to access lat, lon, icon class and popup content.

Use name and type of the facility as content.

screen shot 2016-08-01 at 12 16 25 pm

### DIFF
--- a/src/leaflet/core.cljs
+++ b/src/leaflet/core.cljs
@@ -46,9 +46,17 @@
           new-objects)))))
 
 
-(defn create-marker [[lat lon]]
-  (let [latLng (.latLng js/L lat lon)]
-    (.marker js/L latLng #js {:clickable false})))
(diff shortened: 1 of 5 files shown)


src/leaflet/core.cljs
```diff
@@ -65,8 +73,9 @@
 
 (defmethod leaflet-layer :marker-layer [[_ props & children]]
   (let [layer (.layerGroup js/L)
-        points (:points props)]
-    (doseq [point points] (.addLayer layer (create-marker point)))
+        points (:points props)
+        attrs (select-keys props [:lat-fn :lon-fn :icon-fn :popup-fn])]
+    (doseq [point points] (.addLayer layer (create-marker point props)))
     layer))
```
ggiraldez: This should be `(create-marker point attrs)` right?

spalladino: Fixed

PR #115, opened 2016-08-02
Refactor/cljs

ggiraldez:
Comments/suggestions welcome.


spalladino:
I'm not sure if I like the `components` namespace for react view components, since we also have server-side `components` from stuartsierra. But I don't have a better idea...

PR #130, opened 2016-08-04
Implement project ownership

### LINKED ISSUE #129: Implement project ownership
Part of multi-tenancy support #125 

- Projects should be owned by a user
- Only the owner of a project can access the project (sharing to be implemented later)

ggiraldez:
- Add owner_id column to projects.
- Add user-id in the user identity that is transferred in the session and auth tokens.
- Projects listing will return projects accessible for the user (currently only the ones owned by her).
- The other operations will return a 404 when the user has no permission to access the project.

Closes #129

### DIFF
--- a/src/planwise/component/projects.clj
+++ b/src/planwise/component/projects.clj
@@ -21,8 +21,6 @@
   de application"
   [record]
   (-> record
-      (set/rename-keys {:region_id :region-id
-                        :region_name :region-name})
       (update :stats edn/read-string)
       (update :filters edn/read-string)))
 
@@ -49,8 +47,13 @@
 ;; ----------------------------------------------------------------------
 ;; Service functions
 
-(defn list-projects [service]
+(defn list-projects
+  [service]
   (->> (select-projects (get-db service))
+       (map db->project)))
+
+(defn list-projects-for-user [service user-id]
+  (->> (select-projects-for-user (get-db service) {:user-id user-id})
        (map db->project)))
 
 (defn get-project [service id]
@@ -118,3 +121,13 @@
 
 (defn delete-project [service id]
   (pos? (delete-project* (get-db service) {:id id})))
+
+
+(defn owned-by?
+  [project user-id]
+  (= user-id (:owner-id project)))
+
+(defn accessible-by?
+  [project user-id]
+  (or (owned-by? project user-id)))
+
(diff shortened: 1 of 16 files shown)


src/planwise/component/projects.clj
```diff
@@ -118,3 +121,13 @@
...
   (pos? (delete-project* (get-db service) {:id id})))
+
+
+(defn owned-by?
+  [project user-id]
+  (= user-id (:owner-id project)))
+
+(defn accessible-by?
+  [project user-id]
+  (or (owned-by? project user-id)))
```
spalladino: or... ?

src/planwise/component/projects.clj
```diff
@@ -49,10 +47,15 @@
 ;; ----------------------------------------------------------------------
 ;; Service functions
 
-(defn list-projects [service]
+(defn list-projects
+  [service]
```
spalladino: We should agree (later) on a style guide for this kind of details

src/planwise/component/projects.clj
ggiraldez: Future proofing :)

The complete line should read something like

```clojure
(or (owned-by? project user-id)
    (shared-to? project user-id))
```

src/planwise/component/projects.clj
ggiraldez: Yes we should. I've been using this style lately because it accomodates better to doc-strings. That and it's more uniform with the case of multiple arities.

test/planwise/component/users_test.clj
```diff
@@ -0,0 +1,90 @@
...
+      (is (nil? (:last-login user))))))
+
+(deftest find-user-or-create-test
+  (with-system (system)
+    (let [store (:users-store system)]
+      (let [existing (users/find-or-create-user-by-email store test-email)]
+        (is (some? existing))
+        (is (= test-user-id (:id existing)))
+        (is (= test-full-name (:full-name existing))))
+      (let [_ (execute-sql system "ALTER SEQUENCE users_id_seq RESTART WITH 2")
```
spalladino: What if we change `fixtures` to restart all sequences as well as deleting all data on teardown?

spalladino:
I'd squas the commits (since there are a few WIPs) before merging

test/planwise/component/users_test.clj
ggiraldez: That'd be very nice. We'll need to write our own fixtures adapter though: https://github.com/banzai-inc/fixtures-component#build-your-own-adapter

PR #131, opened 2016-08-04
instedd/planwise#101 add facilities to their isochrones

### LINKED ISSUE #101: Occasionally the facility location is outside the polygon
Found in version 6ec5b5c

Actual result:
Occasionally the facility location is outside the polygon

Expected result:
The facility location should be inside the polygon

Screenshot:
screen shot 2016-07-29 at 2 45 22 pm
screen shot 2016-07-29 at 2 47 18 pm

fgasperij:
Fixes #101.


ggiraldez:
Hmm... we're not using migrations to control SQL functions anymore. See https://github.com/instedd/planwise/tree/master/resources/planwise/plpgsql instead. Split new functions in separate files as necessary. Also, `calculate_isochrones` is deprecated. The function that the importer uses is `process_facility_isochrones` and works on a single facility. We should change `calculate_isochrones` to use it instead.

PR #132, opened 2016-08-04
Add Plumatic Schema

ggiraldez:
Basic schemas defined server side. Enabled validation both client and server side during development.


spalladino:
Wonderful. We can merge as soon as the previous PR is merged.

PR #138, opened 2016-08-09
Calculate unsatisfied demand

### LINKED ISSUE #105: Compute unsatisfied demand raster layer
- Start from the demographics layer
- Order the filtered facilities **by capacity in descending order**
- For each facilitiy, remove the capacity from the unsatisfied demand layer in the area covered by the facility's isochrone. The capacity used for each cell in the isochrone should be proportional to the unsatisfied demand on that cell (wrt. all the isochrone)

spalladino:
Fixes #105


cpp/Makefile
```diff
@@ -0,0 +1,10 @@
+CFLAGS=`gdal-config --cflags`
```
ggiraldez: For production, we want to compile with optimizations turned on `-O3`.

Also, it would be convenient to have warnings enabled, ie. `-Wall`

cpp/calculate-demand.cpp
```diff
@@ -0,0 +1,269 @@
...
+template <typename T> std::string numberToString (T number) {
+  std::ostringstream stringStream;
+  stringStream << number;
+  return stringStream.str();
+}
+
+bool fileExists (const std::string& name) {
+  std::ifstream fileStream(name.c_str());
+  return fileStream.good();
+}
```
ggiraldez: A `stat()` or using boost would be better here.

cpp/calculate-demand.cpp
```diff
@@ -0,0 +1,269 @@
...
+    exit(1);
+  }
+  closeRaster(targetDataset);
+  return atol(metadataValue);
+}
+
+long calculateUnsatisfiedDemand(std::string targetFilename, std::string demoFilename, std::vector<std::string> facilities, std::vector<float> capacities) {
+  GDALDataset* demoDataset = openRaster(demoFilename);
+  GDALRasterBand* demoBand = demoDataset->GetRasterBand(1);
+  CPLAssert(demoBand->GetRasterDataType() == GDT_Float32);
```
ggiraldez: This macro only works in debug mode (ie. when `DEBUG` is defined)

cpp/calculate-demand.cpp
```diff
@@ -0,0 +1,269 @@
...
+      }
+    }
+
+    // TODO: The actual capacity considered should be proportional to the
+    // area of the isochrone that lands on this region
+    float capacity = capacities[iFacility];
+
+    // Each pixel under the isochrone will be multiplied by the proportion of
+    // the unsatisfied demand that is satisfied by this facility.
+    float factor = 1 - (capacity / unsatisfiedCount);
```
ggiraldez: This breaks if `unsatisfiedCount` is zero.

cpp/calculate-demand.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  if (fileExists(argv[1])) {
+#ifdef DEBUG
+    std::cerr << "File " << argv[1] << " already exists." << std::endl;
+#endif
+    unsatisifiedDemand = readUnsatisfiedDemand(argv[1]);
+  } else {
+    std::vector<std::string> facilities;
+    std::vector<float> capacities;
+    for (int i = 3; i < argc; i++) {
```
ggiraldez: Check for an even number of arguments.

cpp/calculate-demand.cpp
```diff
@@ -0,0 +1,269 @@
...
+  std::cerr << " xSize " << xSize << std::endl;
+  std::cerr << " ySize " << ySize << std::endl;
+  std::cerr << " xBlockSize " << xBlockSize << std::endl;
+  std::cerr << " yBlockSize " << yBlockSize << std::endl;
+  std::cerr << " nYBlocks " << nYBlocks << std::endl;
+  std::cerr << " nXBlocks " << nXBlocks << std::endl;
+#endif
+
+  float* buffer = (float*) CPLMalloc(sizeof(float)*xBlockSize*yBlockSize);
+  BYTE* facilityBuffer = (BYTE*) CPLMalloc(sizeof(BYTE)*xBlockSize*yBlockSize);
```
ggiraldez: These are never `CPLFree`'d

cpp/calculate-demand.cpp
spalladino: Should we use plain `assert`, leave it just for development, or use a custom function to check for the condition and `exit(1)` if it does not hold?

src/leaflet/core.cljs
```diff
@@ -101,9 +101,11 @@
     layer))
 
 (defmethod leaflet-layer :tile-layer [[_ props & children]]
-  (let [url (:url props)
-        attrs (dissoc props :url)
-        layer (.tileLayer js/L url (clj->js attrs))]
+  (let [url   (:url props)
+        wms?  (:wms props)
+        attrs (dissoc props :url :wms)
+        constructor (if wms? js/L.tileLayer.wms js/L.tileLayer)
```
ggiraldez: Why not define a `:wms-tile-layer` instead?

cpp/calculate-demand.cpp
spalladino: Fixed in e46864d461ec44cc35db16a996970ca4e6d5bdcf

cpp/calculate-demand.cpp
spalladino: Fixed in e46864d461ec44cc35db16a996970ca4e6d5bdcf

cpp/calculate-demand.cpp
spalladino: Fixed in e46864d461ec44cc35db16a996970ca4e6d5bdcf

cpp/Makefile
spalladino: Fixed in e46864d461ec44cc35db16a996970ca4e6d5bdcf

src/leaflet/core.cljs
spalladino: Fixed in 57d4d87e80720bce28fb661c133357a69326fb4b

src/planwise/component/runner.clj
```diff
@@ -0,0 +1,27 @@
...
+  (apply str (kind service) name))
+
+(defn run-external
+  [service kind name & args]
+  (let [sh-args  (cons (path-for service kind name) args)
+        _        (info "Invoking" (str/join " " sh-args))
+        response (apply sh sh-args)]
+    (case (:exit response)
+      0 (:out response)
+      (throw (RuntimeException. (str "Error running `" (str/join " " sh-args) "`:\n" (:err response)))))))
```
ggiraldez: I am a bit worried that this function doesn't have timeout control of some sort, especially since it can be invoked in the context of a HTTP request.

src/planwise/component/maps.clj
```diff
@@ -44,26 +48,28 @@
...
-                    (vec polygons-with-capacities))
-        unsatisfied-count (-> response
-                            (str/trim-newline)
-                            (str/trim)
-                            (Integer.))]
-      {:map-key map-key,
-       :unsatisfied-count unsatisfied-count}))
+  (if-not (calculate-demand? service)
+    {}
+    (let [polygons (filter :polygon-id facilities)
```
ggiraldez: `(when (calculate-demand? service) ...)` is a more idiomatic equivalent

src/planwise/component/runner.clj
```diff
@@ -8,16 +8,23 @@
...
+    (throw (ex-info (str "Path for " kind " not found") {:kind kind}))))
 
 (defn run-external
-  [service kind name & args]
+  [service kind timeout name & args]
   (let [sh-args  (cons (path-for service kind name) args)
         _        (info "Invoking" (str/join " " sh-args))
-        response (apply sh sh-args)]
+        response (if timeout
+                   (deref (future (apply sh sh-args)) timeout {:exit :timeout, :err :timeout})
```
ggiraldez: 👏

spalladino: 😄

src/planwise/component/maps.clj
spalladino: Thing is I wanted to return an empty hash, instead of nil

src/planwise/component/facilities.clj
```diff
@@ -93,28 +96,36 @@
...
-        (set-facility-polygon-region-population!
-          (get-db service)
-          (assoc fpr :population population))))))
+      (try
+        (let [population (-> service
+                            (:runner)
+                            (run-external :scripts 60000 "raster-isochrone" (str region-id) (str facility-polygon-id))
+                            (trim-newline)
+                            (trim)
+                            (Integer.))]
```
ggiraldez: While this works fine, conceptually it mixes two different "pipelines": command construction and output processing. I think
```clojure
(-> (run-external (:runner service) :scripts 60000 ...)
    trim-newline
    trim
    Integer.)
```
has clearer intent (btw, the parens are optional when it's a single element)
Also, the `(-> x trim-newline trim Integer.)` pattern is used 3 times. Care to refactor in a util function? We probably need a general util namespace.

src/planwise/component/maps.clj
ggiraldez: Why? `nil` is equivalent in 99% of the cases (only exception I can think of is `map?`). The only use I could find of this function is immediately followed by an `assoc` which works as intended with `nil`.

src/planwise/component/facilities.clj
spalladino: Fixed in ea8605f5bb469be3541bcca9de98b6ab4ed53f9d

src/planwise/component/maps.clj
spalladino: True, as always. Fixed in ea8605f5bb469be3541bcca9de98b6ab4ed53f9d.

cpp/calculate-demand.cpp
ggiraldez: It depends on how catastrophic is that the assertion is not valid. For user input validation (as in program parameters in this case) aborting with `exit(1)` seems the right choice here.

PR #141, opened 2016-08-10
Docker compose

spalladino:
Set up docker compose for running in local machine.

1- Create `.docker-env` file with GUISSO env vars
2- Run `docker-compose -f docker-compose.setup.yml up` to run import scripts
3- After `planwise_setup_1` exits successfully, run `docker-compose up` to start the app in port 3000


resources/migrations/015-apply_traffic_factor.up.sql
```diff
@@ -1,10 +1,8 @@
...
-UPDATE ways
-       SET cost_s = length_m / (maxspeed_forward * 5 / 18) * 1.5;
+DO $$
+  BEGIN
+    IF EXISTS (SELECT 1 FROM pg_class WHERE relname = 'ways') THEN
+      UPDATE ways
+      SET cost_s = length_m / (maxspeed_forward * 5 / 18) * 1.5;
+    END IF;
+  END;
+$$
```
ggiraldez: Why wrap it in a `DO ... END` here but not for the .down file?

spalladino: I thought it was obvious: because I absolutely forgot about the .down file.
Fixed in 350b21e9ab1764a22260a716ce1503960f56bbc1.

PR #143, opened 2016-08-11
instedd/planwise align crop isochrones to region blocks.

fgasperij:


cpp/calculate-demand.cpp
```diff
@@ -72,68 +73,88 @@ long readUnsatisfiedDemand(std::string targetFilename) {
...
+    // open isochrone dataset
     GDALDataset* facilityDataset = openRaster(facilities[iFacility]);
     GDALRasterBand* facilityBand = facilityDataset->GetRasterBand(1);
     BYTE facilityNodata = facilityBand->GetNoDataValue();
-
     CPLAssert(facilityBand->GetRasterDataType() == GDT_Byte);
-    int xBlockSizeFacility, yBlockSizeFacility;
-    facilityBand->GetBlockSize(&xBlockSizeFacility, &yBlockSizeFacility);
-    CPLAssert(xBlockSizeFacility == xBlockSize);
-    CPLAssert(yBlockSizeFacility == yBlockSize);
```
spalladino: Let's keep the assertions on blocksize, as they should still match between all rasters

cpp/calculate-demand.cpp
```diff
@@ -72,68 +73,88 @@ long readUnsatisfiedDemand(std::string targetFilename) {
...
+    // set blocks offsets
+    double targetProjection[6], facilityProjection[6];
+    targetDataset->GetGeoTransform(targetProjection);
+    facilityDataset->GetGeoTransform(facilityProjection);
+
+    double facilityMaxY = facilityProjection[3];
+    double demoMaxY = facilityProjection[3];
+    // double facilityYRes = facilityProjection[5];
+    double demoYRes = targetProjection[5];
+    // TODO: assert that both res are equal
```
spalladino: Add the actual assertion here ;)

cpp/calculate-demand.cpp
```diff
@@ -72,68 +73,88 @@ long readUnsatisfiedDemand(std::string targetFilename) {
...
+    int blocksRowOffset = std::abs(round( (facilityMaxY-demoMaxY)/(128 * demoYRes) ));
+
+    double facilityMinX = facilityProjection[0];
+    double demoMinX = facilityProjection[0];
+    // double facilityXRes = facilityProjection[1];
+    double demoXRes = targetProjection[1];
+    // TODO: assert that both res are equal
```
spalladino: Same as previous one

cpp/calculate-demand.cpp
```diff
@@ -72,68 +73,88 @@ long readUnsatisfiedDemand(std::string targetFilename) {
...
+    double targetProjection[6], facilityProjection[6];
+    targetDataset->GetGeoTransform(targetProjection);
+    facilityDataset->GetGeoTransform(facilityProjection);
+
+    double facilityMaxY = facilityProjection[3];
+    double demoMaxY = facilityProjection[3];
+    // double facilityYRes = facilityProjection[5];
+    double demoYRes = targetProjection[5];
+    // TODO: assert that both res are equal
+    int blocksRowOffset = std::abs(round( (facilityMaxY-demoMaxY)/(128 * demoYRes) ));
```
spalladino: Shouldn't it be `floor` instead of `round` here? Also: is the `abs` needed? Shouldn't the maxY of the demo layer be always greater than the facility's?

cpp/calculate-demand.cpp
```diff
@@ -162,17 +183,17 @@ long calculateUnsatisfiedDemand(std::string targetFilename, std::string demoFile
 #endif
 
     // Second pass: we apply the multiplying factor to each pixel
-    for (int iXBlock = 0; iXBlock < nXBlocks; ++iXBlock) {
+    for (int iXBlock = 0; iXBlock < facilityNXBlocks; ++iXBlock) {
       xOffset = iXBlock*xBlockSize;
       nXValid = xBlockSize;
-      if (iXBlock == nXBlocks-1) nXValid = xSize - xOffset;
+      if (iXBlock == facilityNXBlocks-1) nXValid = facilityXSize - xOffset;
```
spalladino: Can we add an assertion here to double-check that we don't need to check the previous condition (ie `xSize - xOffset` when on the last x block of the region raster)? Since blocks are aligned and the facility's raster should be contained within the region's one, it should never happen, but better be safe than sorry.

scripts/calculate_aligned_extent.sql
```diff
@@ -0,0 +1,59 @@
...
+  from regions
+  where id = region_id;
+
+  -- calculate extent based on the block size of (128x128) and the precision of 0.0008333
+  unaligned_target_minx := greatest(isochrone_xmin, region_xmin);
+  unaligned_target_miny := greatest(isochrone_ymin, region_ymin);
+  unaligned_target_maxx := least(isochrone_xmax, region_xmax);
+  unaligned_target_maxy := least(isochrone_ymax, region_ymax);
+
+  srs_block_length := 0.000833*128;
```
spalladino: Avoid hardcoding these values here. We should extract them from the raster and pass them as extra params to the function. `gdalinfo` will return this information, and can be extracted using a regex (FWIW, python, grep, sed, perl and awk are all available on the app docker image); or we can write a small cpp that uses the GDAL API to get exactly those values.

scripts/calculate_aligned_extent.sql
```diff
@@ -0,0 +1,59 @@
+create or replace function calculate_aligned_extent(facility_polygon_id integer, region_id integer)
```
spalladino: We are redefining the function on every invocation of the script here. Move this to `resources/planwise/sql/` folder so the function is set up via migrations, and invoke it directly from the bash script.

cpp/calculate-demand.cpp
fgasperij: This assertions aren't working, we could replace them with custom asserts. Would that be OK?

cpp/calculate-demand.cpp
fgasperij: It was an attempt to prevent errors that could arose from floating point math when dealing with regions and isochrones with similar bounds. The abs and round are interrelated in this case because floor takes a negative number to it's previous integer which is greater than the searched value.  We could assume we will never get negative numbers there and use floor without abs but I suggest at least adding an assertion of that fact. What do you think?

cpp/calculate-demand.cpp
fgasperij: Good catch. Yes, I think it will be good to add that check but outside the loop so we check it just once and the error, if it occurs, it does in a more reasonable place. Something like: `assert(facilitySize < (demoSize - (blockSize * blocksOffset))`

cpp/calculate-demand.cpp
spalladino: It seems that they only work if DEBUG is defined (see comment by ggiraldez [here](https://github.com/instedd/planwise/pull/138#discussion_r74132046)). Feel free to change them to custom asserts if you want.

spalladino:
@fgasperij let's delete `scripts/calculate_aligned_extent.sql` and we're good to go

Issue #106, opened 2016-08-01
Precalculate population on each region


ggiraldez:
Using the demographics layer and the region's boundary, compute the total population living in the region's area and cache it in the regions table.

Re-enable the demographics tab in the project view and show the population in the project's region.

mmuller:
Ok in version v0.4.0-151-gf78af9f

PR #157, opened 2016-08-17
Display region population for projects

juanedi:
See #106

### DIFF
--- a/src/planwise/client/projects/components/sidebar.cljs
+++ b/src/planwise/client/projects/components/sidebar.cljs
@@ -6,10 +6,25 @@
             [planwise.client.projects.db :as db]))
 
 
+(defn- demographic-stat [title value]
+  [:div.stat
+   [:div.stat-title title]
+   [:div.stat-value value]])
+
 (defn- demographics-filters []
-  [:div.sidebar-filters
-   [:div.filter-info
-    [:p "Filter here the population you are analyzing."]]])
+  (let [current-project (subscribe [:projects/current-data]) ]
+    (fn []
+      [:div.sidebar-filters
+      [:div.filter-info
+       ;; [:p "Filter here the population you are analyzing."]
+       [:div.demographic-stats
+        ;; (demographic-stat "Area" (str 1000 " km2"))
+        ;; (demographic-stat "Density" (str 4850 "/km2"))
+        (demographic-stat "Total population" (.toLocaleString (:region-population @current-project) "en"))
+       ]
+       [:span.small
+        "Information source: "
+        [:a {:href "http://www.worldpop.org.uk/"} "WorldPop"]]]])))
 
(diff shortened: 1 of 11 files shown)


src/planwise/client/projects/components/sidebar.cljs
```diff
@@ -1,15 +1,35 @@
...
+       [:div.filter-info
+        ;; [:p "Filter here the population you are analyzing."]
+        [:div.demographic-stats
+         (demographic-stat "Area" [:span (utils/format (int @area)) " km" [:sup 2]])
+         (demographic-stat "Density" [:span (utils/format @density) " /km" [:sup 2]])
+         (demographic-stat "Total population" (utils/format @population))
+         ]
+        [:span.small
+         "Population information source: "
+         [:a {:href "http://www.worldpop.org.uk/"} "WorldPop"]]]])))
```
spalladino: Let's use this space to provide proper attribution according to [CCv4 license](https://creativecommons.org/licenses/by/4.0/). We'll probably need to cite also the author (Geodata institute of the university of Southampton), see if there is a suggested way to cite them in the worldpop site.

Also, let's change _population information source_ to _demographic data source_ to be consistent with the wording of the section.

src/planwise/client/projects/components/sidebar.cljs
```diff
@@ -1,15 +1,35 @@
...
+   [:div.stat-value value]])
+
 (defn- demographics-filters []
-  [:div.sidebar-filters
-   [:div.filter-info
-    [:p "Filter here the population you are analyzing."]]])
+  (let [current-project (subscribe [:projects/current-data])
+        population (reaction (:region-population @current-project))
+        area (reaction (:region-area-km2 @current-project))
+        density (reaction (/ @population @area))]
```
spalladino: These 3 don't need to be a reaction, since they will not be changing independently of the current project. Remove the `reaction`, and remove the deref (ie the `@`) from their uses in the code.

test/planwise/component/projects_test.clj
```diff
@@ -0,0 +1,44 @@
...
+  (with-system (system)
+    (let [service (:projects system)
+          project (projects/get-project service 1)]
+      (is (= 1000 (:region-population project))))))
+
+(deftest region-area-is-retrieved-on-get-project
+  (with-system (system)
+    (let [service (:projects system)
+          project (projects/get-project service 1)]
+      (is (< 0 (:region-area-km2 project))))))
```
spalladino: `pos?` is more idiomatic ;-)

spalladino: Tests here do not finish on a failed assertion, so you should be able to compact `region-population-is-retrieved-on-get-project` and `region-area-is-retrieved-on-get-project` into a single test with both assertions together.

PR #158, opened 2016-08-17
Importer refactor

### LINKED ISSUE #94: Datasets information is not valid if page refreshed while importing
Steps to reproduce:

1. Start an import process
2. Navigate to projects
3. Reload the page
4. Wait for a status tick to request the importer status
5. Navigate back to datasets

Obtained: the page incorrectly shows that the user is not authorized to access Resourcemap.
Expected: the collection list is shown (disabled) along the import status.

This is because the status tick overwrites the `[:datasets :state]` which is initially nil, but not the rest of the information. Then later the client assumes it has already initialized the authorization/collections and doesn't issue the initial request to the datasets endpoint.

### LINKED ISSUE #120: Refactor and extend importer component
The current implementation is severely limited in functionality and scalability.

Reimplement and do one control thread, constantly listening on the control channel, and several workers threads. Implement the capability of spawning several simultaneous jobs.

Also add persistency to the executing jobs and allow resuming operation after a service restart.

ggiraldez:
Fixes #94. Closes #120.

Complete reimplementation of the importer component.

- The import job itself is now modelled with a FSM supporting cancellation in all stages.
- There's a TaskMaster asynchronous component which dispatches a configurable number of parallel tasks on demand.
- The job FSM handles task dispatching depending on the stage the import process is in.
- Fix import status handling on the client side as well, improving concurrency issues and closing #94.
- Progress report has been improved and will now report Resourcemap import progress as well as facility isochrones processing.
- Project updating is done as soon as the sites have been imported to improve project stats consistency during the import process.
- Last import operation result is shown in the UI.
- Other minor fixes and changes.

import-job-fsm

### DIFF
--- a/dev/dev.clj
+++ b/dev/dev.clj
@@ -34,8 +34,7 @@
    :paths {:bin "cpp/"
            :scripts "scripts/"
            :data "data/"}
-   :maps {:mapserver-url "http://localhost:5002/mapcache?"
-          :facilities-capacity 2000000}
+   :maps {:facilities-capacity 2000000}
    :figwheel
    {:css-dirs ["resources/planwise/public/css"
                "target/sass-repl"]
--- a/project.clj
+++ b/project.clj
@@ -43,6 +43,7 @@
                  [hiccup "1.0.5"]
                  [cheshire "5.6.3"]
                  [clj-time "0.12.0"]
+                 [reduce-fsm "0.1.4"]
 
                  ; Client infrastructure
                  [reagent "0.5.1"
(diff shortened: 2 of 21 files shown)


spalladino:
On the FSM flow: if a `cancel` is issued while on processing facilities, the clean up task will dispatch an `update-projects` job, which had already been invoked during the `update-projects` stage (between importing sites and processing facilities). Is it necessary to run it again in this case?

ggiraldez:
Good catch. That's a left over from when update projects was done as the last stage. It's not necessary (at least for now). I'll fix and update the PR.

PR #159, opened 2016-08-17
Load geojson by bounding box

### LINKED ISSUE #78: Tweak the simplify parameter when retrieving isochrones
If we continue to use vector isochrones (returned as GeoJSON) we'll need to simplify the returned result. Currently, with ~15 facilities and a threshold of 2 hours the resulting request is 370 Kb. This will not scale with a larger number of facilities.

See also #77.

### LINKED ISSUE #139: Cache isochrones client-side
So that switching between different transport times doesn't reload every isochrone each time.

spalladino:
Add bbox-loader leaflet component that manages the features of a layer group (in this case, a geojson group layer) by requesting them by bounding box and level, from a supplied callback. The loader keeps track of the _tiles_ already requested, to minimise the number of requests. The _level_ is inferred from the current map zoom level, and it is responsibility of its client to decide the resolution of the features provided by level.

Additionally, the client _db_ keeps track of the isochrones loaded by threshold, level and facility-id. When the bbox-loader requests the features for a bbox, it will make a request to a new endpoint on the server, with the bbox and the ids of the isochrones already loaded. The server will answer with all isochrones in the bbox that are not in the set of already loaded ids, so as to reduce bandwidth.

This has the side effect of caching the isochrones for different thresholds in the client _db_, so switching back to a threshold within a project will not re-download all isochrones from the server (it will make a request though, answered by the server with the list of facilities, yet with no isochrones).

As a further optimisation, the client could also keep track of the features by bounding box, or the set of tiles already requested, and only make the request to the server if there are new tiles needed.

Fixes #139 
Fixes #78

### DIFF
--- a/resources/planwise/sql/facilities.sql
+++ b/resources/planwise/sql/facilities.sql
@@ -54,6 +54,25 @@ FROM facilities
   /*~ ) ~*/
 WHERE 1=1 :snip:criteria ;
 
+-- :name isochrones-in-bbox* :?
+SELECT
+  facilities.id AS "id",
+  fp.id AS "polygon-id",
+  /*~ (if (empty? (:excluding params)) */
+  ST_AsGeoJSON(ST_Simplify(fp.the_geom, :simplify)) AS "isochrone"
+  /*~*/
+  CASE facilities.id IN (:v*:excluding)
+    WHEN FALSE THEN ST_AsGeoJSON(ST_Simplify(fp.the_geom, :simplify))
+    ELSE NULL
+  END AS "isochrone"
+  /*~ ) ~*/
+FROM facilities
+  INNER JOIN facility_types ON facilities.type_id = facility_types.id
+  LEFT OUTER JOIN facilities_polygons fp ON fp.facility_id = facilities.id AND fp.threshold = :threshold AND fp.method = :algorithm
+WHERE
+  ST_Intersects(fp.the_geom, ST_MakeEnvelope(:v*:bbox, 4326))
+  :snip:criteria ;
+
 -- :name isochrone-for-facilities :? :1
 SELECT
   ST_AsGeoJSON(ST_Union(the_geom))
--- a/resources/planwise/sql/facilities.sql
+++ b/resources/planwise/sql/facilities.sql
@@ -70,7 +70,7 @@ FROM facilities
   INNER JOIN facility_types ON facilities.type_id = facility_types.id
   LEFT OUTER JOIN facilities_polygons fp ON fp.facility_id = facilities.id AND fp.threshold = :threshold AND fp.method = :algorithm
 WHERE
-  ST_Intersects(fp.the_geom, ST_MakeEnvelope(:v*:bbox, 4326))
+  (fp.the_geom && ST_MakeEnvelope(:v*:bbox, 4326))
   :snip:criteria ;
 
 -- :name isochrone-for-facilities :? :1
--- a/src/planwise/client/projects/handlers.cljs
+++ b/src/planwise/client/projects/handlers.cljs
@@ -169,7 +169,7 @@
    (let [level (maps/simplify->geojson-level simplify)
          isochrones  (->> facilities
                        (filter #(some? (:isochrone %)))
-                       (map (juxt :id #(js/JSON.parse (:isochrone %))))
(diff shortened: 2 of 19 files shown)


spalladino:
Pending: 
- [ ] Review simplify factors, tile sizes and zoom levels.
- [ ] Fix _request uri too long_ error when working with large number of facilities

resources/planwise/sql/facilities.sql
```diff
@@ -54,6 +54,25 @@ FROM facilities
...
+  CASE facilities.id IN (:v*:excluding)
+    WHEN FALSE THEN ST_AsGeoJSON(ST_Simplify(fp.the_geom, :simplify))
+    ELSE NULL
+  END AS "isochrone"
+  /*~ ) ~*/
+FROM facilities
+  INNER JOIN facility_types ON facilities.type_id = facility_types.id
+  LEFT OUTER JOIN facilities_polygons fp ON fp.facility_id = facilities.id AND fp.threshold = :threshold AND fp.method = :algorithm
+WHERE
+  ST_Intersects(fp.the_geom, ST_MakeEnvelope(:v*:bbox, 4326))
```
ggiraldez: Is there any reason for not using the `&&` operator instead of `ST_Intersects`. AFAIK, `&&` will use the index whereas `ST_Intersects` forces a sequential scan.

src/planwise/client/projects/views.cljs
```diff
@@ -16,6 +20,29 @@
...
+        (fn [response]
+          (dispatch-sync [:projects/isochrones-loaded response])
+          (let [isochrones-for-level (get-in @isochrones [threshold level])
+                new-isochrones (->> response
+                                  (:facilities)
+                                  (map :id)
+                                  (remove (set bbox-excluding))
+                                  (select-keys isochrones-for-level)
+                                  (map (fn [[id isochrone]] {:id id, :isochrone isochrone}))
+                                  (vec))]
```
ggiraldez: Use `(mapv ...)` instead?

src/planwise/endpoint/facilities.clj
```diff
@@ -2,18 +2,27 @@
...
+            [clojure.string :as string]
             [buddy.auth.accessrules :refer [restrict]]
             [buddy.auth :refer [authenticated?]]
             [ring.util.response :refer [response]]))
 
-(defn facilities-criteria [type region]
-  {:region (if region (Integer. region) nil)
-   :types (if type (map #(Integer. %) (vals type)) nil)})
+(defn- facilities-criteria [{:keys [type region bbox excluding]}]
+  {:region (when region (Integer. region))
```
ggiraldez: Alternatively, you can use `(some-> region Integer.)` which *may* be more readable/idiomatic

src/planwise/endpoint/facilities.clj
```diff
@@ -2,18 +2,27 @@
...
+   :types (when type (map #(Integer. %) (vals type)))
```
ggiraldez: Ditto with `(some->> type vals (map #(Integer/parseInt %)))`

src/planwise/client/projects/views.cljs
```diff
@@ -16,6 +20,29 @@
             [planwise.client.projects.components.sidebar
              :refer [sidebar-section]]))
 
+(defn geojson-bbox-callback [isochrones filters threshold]
```
ggiraldez: This makes an API call in the `views` namespace which seems a bit out of place, but it's true that there's no other clear place to put it in. Maybe as a subscription?

src/planwise/client/projects/handlers.cljs
```diff
@@ -169,26 +163,19 @@
...
-     (assoc-in [:facilities :isochrones] (->> response
-                                           :facilities
-                                           (map :isochrone)
-                                           (set)
-                                           (filterv some?))))))
+ (fn [db [_ {:keys [map-key unsatisfied-count facilities threshold simplify], :as response}]]
+   (let [level (maps/simplify->geojson-level simplify)
+         isochrones  (->> facilities
+                       (filter #(some? (:isochrone %)))
+                       (map (juxt :id #(js/JSON.parse (:isochrone %))))
```
ggiraldez: `(comp js/JSON.parse :isochrone)` is cooler :P

PR #166, opened 2016-08-18
Fix/project issues

### LINKED ISSUE #153: Remove project search bar when there are no projects
Update the projects list screen for the "no data" case.

ggiraldez:
Fix a couple of (minor) projects issues related to creation and deleting.

Fixes #153.

### DIFF
--- a/resources/planwise/sql/projects.sql
+++ b/resources/planwise/sql/projects.sql
@@ -12,18 +12,33 @@ INNER JOIN regions ON projects.region_id = regions.id
 ORDER BY projects.id ASC;
 
 -- :name select-projects-for-user :?
+-- This query should include all fields necessary for the project list
 SELECT
-  projects.id, projects.goal, projects.region_id AS "region-id", projects.stats,
-  regions.name AS "region-name", regions.total_population AS "region-population", owner_id AS "owner-id"
+  projects.id,
+  projects.goal,
+  projects.region_id AS "region-id",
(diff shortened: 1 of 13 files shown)


spalladino:
Regarding https://github.com/instedd/planwise/pull/166/commits/2ae9775b92bf2f0c16ea21c0fadf2fe2b106a8af, why do we even issue a new `list-projects` request when navigating to the projects list? All projects accessible by a user are those created by him, so there should be no need to actually reload that list; and we can always add a manual "reload" button if we need to work around that.

Another option that comes to mind is, instead of issuing a duplicate request after the delete is confirmed, keep in the `db` a list of projects _deleted_ in the current session, and always filter the project list removing the deleted projects.

ggiraldez:
Good point. Projects being per-user makes things a lot simpler. I don't like the "reload" button though. It's an "unf*ck me" button and we shift responsibility to the user for something we own.

Another option is to make the delete endpoint return the new list, breaking REST semantics, but that I'd be willing to pay. Other ideas?

spalladino:
How about the `deleted-ids` entry in the client `db`, used for filtering in the current session?

ggiraldez:
Works too. For some reason I don't like it as much. If the delete fails we should remove the id from that list.

spalladino:
If the delete fails, we should be doing the same as if any other optimistic update request fails: panic.

PR #168, opened 2016-08-19
Style markers by facility type

### LINKED ISSUE #147: Render facilities with color/symbol depending on the type
From Concern feedback, Taylor:

>Maybe have different colored dots for the different types of centers so you can differentiate between the types of facilities on the map? (for example: hospitals = purple, dispensaries = red, etc.)

Stephen:

>Color/shape filtering for facility type? A hover key for what facility types mean/offer and perhaps identify facilities appropriate to certain issues/conditions? Building in a set of best practice parameters for a range of issues/conditions?

spalladino:
Fixes #147

screen shot 2016-08-19 at 4 40 57 pm

### DIFF
--- a/src/leaflet/core.cljs
+++ b/src/leaflet/core.cljs
@@ -96,9 +96,9 @@
     layer))
 
 (defmethod leaflet-layer :point-layer [[_ props & children]]
-  (let [layer (.layerGroup js/L)
+  (let [layer  (.withPathGroup js/L.layerGroup)
         points (:points props)
-        attrs (dissoc props :points)]
+        attrs  (dissoc props :points)]
     (doseq [point points] (.addLayer layer (create-point point attrs)))
     layer))
 
--- a/src/leaflet/core.cljs
+++ b/src/leaflet/core.cljs
@@ -157,7 +157,10 @@
 (defn leaflet-update-layers [this]
   (let [state (reagent/state this)
         leaflet (:map state)
-        new-children (reagent/children this)]
+        new-children (->> this
+                       (reagent/children)
+                       (mapcat #(if (seq? %) % [%]))
+                       (vec))]
 
     ;; go through all the layers, old and new, and update the Leaflet objects
     ;; accordingly while updating the map
--- a/src/leaflet/core.cljs
+++ b/src/leaflet/core.cljs
@@ -157,10 +157,7 @@
 (defn leaflet-update-layers [this]
   (let [state (reagent/state this)
         leaflet (:map state)
-        new-children (->> this
-                       (reagent/children)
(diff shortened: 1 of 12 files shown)


src/planwise/client/projects/views.cljs
```diff
@@ -92,29 +92,30 @@
...
+                                    :opacity 0.5}])
+                ;; Markers with filtered facilities
+                (when (#{:facilities :transport} selected-tab)
+                  (for [[type facilities] @facilities-by-type]
+                    [:point-layer {:points facilities
+                                   :popup-fn marker-popup-fn
+                                   :radius 4
+                                   :color (:colour type)
+                                   :stroke false
+                                   :fillOpacity 1}]))
```
ggiraldez: I think it's time to refactor this. I'd implement a separate function to yield the map's layers and splat them `into` a base map, also created separately. This will yield the ugly `mapcat` in `leaflet-update-layers` unnecessary.

The result would look something like this:

```clojure
        [:div
         [sidebar-section selected-tab]
         [:div.map-container
          (let [map-props {:position @map-position
                           :zoom @map-zoom
                           :min-zoom 5
                           :on-position-changed
                           #(dispatch [:projects/update-position %])
                           :on-zoom-changed
                           #(dispatch [:projects/update-zoom %])}
                layers [;; Base tile layer
                        mapping/gray-base-tile-layer
                        ;; Demographics tile layer
                        (let [demand-map ...
                              population-map ...]
                          [:wms-tile-layer ...])
                        ;; Boundaries of working region
                        (if @map-geojson
                          [:geojson-layer ...])
                        ;; Markers with filtered facilities
                        (when (#{:facilities :transport} selected-tab)
                          [:point-layer {:points @facilities
                                         :popup-fn marker-popup-fn
                                         :radius 5
                                         :color styles/light-grey
                                         :stroke false
                                         :fillOpacity 1}])
                        ;; Isochrone for selected transport
                        (when (= :transport selected-tab)
                          [:geojson-bbox-layer ...])]]
            (into [map-widget map-props] (filter some? layers)))]]
```

Obviously this is from before this PR, but the idea is the same. Pre-process the children before yielding the Reagent component.

src/leaflet/core.cljs
```diff
@@ -157,7 +157,10 @@
 (defn leaflet-update-layers [this]
   (let [state (reagent/state this)
         leaflet (:map state)
-        new-children (reagent/children this)]
+        new-children (->> this
+                       (reagent/children)
+                       (mapcat #(if (seq? %) % [%]))
+                       (vec))]
```
ggiraldez: This is unnecessary. See comment below.

PR #169, opened 2016-08-22
Add loading signs

### LINKED ISSUE #140: Add loading indicators
There are "Loading..." legends here and there, but we need more, esp. for operations which change the map. Currently updating filters/transport time and loading facilities and isochrones can take a couple of seconds even on fast connections.

fgasperij:
Fixes #140


resources/sass/_projects.scss
```diff
@@ -114,6 +114,66 @@ $search-box-height: 42px;
...
+  bottom: 25px;
+
+  z-index: 10;
+
+  width: 170px;
+  height: 45px;
+  padding-left: 10px;
+  padding-right: 10px;
+
+  background-color: #666666;
```
ggiraldez: Please, use (or define new) colors from `_defs.scss`

resources/sass/_projects.scss
```diff
@@ -114,6 +114,66 @@ $search-box-height: 42px;
...
+  border-radius: 4px;
+
+  font-weight: 500;
```
ggiraldez: Why not `bold`?

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+            (js/setTimeout #(dispatch [:projects/trigger-map-request]) 500)))
+
+(defn cancel-prev-timeout [db]
+  (js/clearTimeout (get-in db [:map-state :timeout]))
+  (assoc-in db [:map-state :timeout] nil))
+
+(defn cancel-prev-request [db]
+  (ajax.protocols/-abort (get-in db [:map-state :request]))
+  (assoc-in db [:map-state :request] nil))
+
```
ggiraldez: Unless you previously `declare` all these, you must define them in reverse-order, ie. must be already defined when you use them.

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+(defn loading-displayed-to-request-pending [db]
+  (-> db (cancel-prev-request) (set-request-timeout)))
+
+(defn set-request-timeout [db]
+  (assoc-in db [:map-state :timeout]
+            (js/setTimeout #(dispatch [:projects/trigger-map-request]) 500)))
+
+(defn cancel-prev-timeout [db]
+  (js/clearTimeout (get-in db [:map-state :timeout]))
+  (assoc-in db [:map-state :timeout] nil))
```
ggiraldez: `(update-in db [:map-state :timeout] js/clearTimeout)` is more readable and does exactly the same here.

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+        (assoc-in [:map-state :request-with] request-with))))
+
+(defn loaded-to-request-pending [db]
+  (set-request-timeout db))
+
+(defn request-pending-to-request-pending [db]
+  (-> db (cancel-prev-timeout) (set-request-timeout)))
+
+(defn loading-to-request-pending [db]
+  (-> db (cancel-prev-timeout) (cancel-prev-request) (set-request-timeout)))
```
ggiraldez: A vertical layout here is more readable. Ie.

```clojure
(-> db
    cancel-prev-timeout
    cancel-prev-request
    set-request-timeout)
```
Also, note parens are optional after the first argument.

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+                   (assoc-in [:transport :time] time)
+                   (transition-map-state nil))]
       new-db)))
 
 (register-handler
  :projects/project-updated
  in-current-project
  (fn [db [_ project]]
-   (db/update-viewmodel db project)))
+   (let [prev-state (get-in % [:map-state :current])
```
ggiraldez: There's a typo here. `%` should read `db`.

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+         new-db (-> db
+                    (#(if (= prev-state :loading) (cancel-prev-timeout %) %))
```
ggiraldez: The conditional is not necessary here. `cancel-prev-timeout` is idempotent.

src/planwise/client/projects/views.cljs
```diff
@@ -58,20 +58,29 @@
...
           [new-project-dialog]])])))
 
+(defn loading-wheel []
+  [:svg.circular {:viewBox "25 25 50 50"}
+   [:circle.path {:cx 50
+                  :cy 50
+                  :fill "none"
+                  :r 20
+                  :stroke-miterlimit 10
+                  :stroke-width 2}]])
```
ggiraldez: Ok, but since it's a constant, it can simply be `def`'ed and referenced directly later.

src/planwise/client/projects/views.cljs
```diff
@@ -80,6 +89,9 @@
          selected-tab)
         [:div
          [sidebar-section selected-tab]
+         [:div.loading-indicator {:style {:display (if (= @map-state :loading-displayed) "block" "none")}}
+          [:div.loading-wheel (loading-wheel)]
```
ggiraldez: Ok since `loading-wheel` is a constant. Otherwise it's advisable to return the function unevaluated as a component. Ie.

```clojure
[loading-wheel]
```
and move the surrounding `div` to the component itself.

src/planwise/client/projects/views.cljs
```diff
@@ -80,6 +89,9 @@
          selected-tab)
         [:div
          [sidebar-section selected-tab]
+         [:div.loading-indicator {:style {:display (if (= @map-state :loading-displayed) "block" "none")}}
```
ggiraldez: This is an incorrect use of React's model. You should just:

```clojure
(when (= @map-state :loading-displayed)
   [:div.loading-indicator
     [loading-wheel]
     [:span "Retrieving facilities"]])
```

The component should not be rendered (ie. not hidden) if the condition is not met.

resources/sass/_projects.scss
fgasperij: Because bold is 700, not 500.

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+   (let [filters (db/project-filters db)
+         project-id (get-in db [:project-data :id])
+         request-with (get-in db [:map-state :request-with])]
+     (-> db
+         (assoc-in [:map-state :request] (api/update-project project-id filters request-with :projects/project-updated))
+         (assoc-in [:map-state :timeout] (js/setTimeout #(dispatch [:projects/show-map-loading-sign]) 2500))
+         (assoc-in [:map-state :current] :loading)))))
+
+(register-handler
+ :projects/show-map-loading-sign
```
ggiraldez: Not a fan of the name `sign`

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
                     (assoc-in [:facilities :isochrones] nil))
          filters (db/project-filters new-db)
          project-id (get-in db [:project-data :id])]
-     (api/update-project project-id filters :facilities
-                         :projects/project-updated)
-     new-db)))
+     (transition-map-state new-db :facilities))))
+
+(defn transition-map-state [db request-with]
```
ggiraldez: This name is not clear as to what's the intention of the function. Maybe `start-update-request` or something along those lines?

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+(defn set-request-timeout [db]
+  (assoc-in db [:map-state :timeout]
+            (js/setTimeout #(dispatch [:projects/trigger-map-request]) 500)))
+
+(defn cancel-prev-timeout [db]
+  (js/clearTimeout (get-in db [:map-state :timeout]))
+  (assoc-in db [:map-state :timeout] nil))
+
+(defn cancel-prev-request [db]
+  (ajax.protocols/-abort (get-in db [:map-state :request]))
```
ggiraldez: Careful here. If the request is `nil` this crashes.

src/planwise/client/projects/handlers.cljs
```diff
@@ -194,26 +194,81 @@
...
+(register-handler
+ :projects/trigger-map-request
+ in-current-project
+ (fn [db _]
+   (let [filters (db/project-filters db)
+         project-id (get-in db [:project-data :id])
+         request-with (get-in db [:map-state :request-with])]
+     (-> db
+         (assoc-in [:map-state :request] (api/update-project project-id filters request-with :projects/project-updated))
+         (assoc-in [:map-state :timeout] (js/setTimeout #(dispatch [:projects/show-map-loading-sign]) 2500))
```
ggiraldez: Please move the magic constants as `def`s at the beginning of the namespace.

src/planwise/client/projects/handlers.cljs
fgasperij: Ok. Is this a style guideline? It works as it is.

ggiraldez: Actually, it compiles with warnings because the target is JS. This doesn't compile at all in Clojure.

resources/sass/_projects.scss
ggiraldez: Fair enough, can you define a `$semi-bold` constant in `_defs.scss` and use that instead then? We need to review font styles with @jkicillof at some point anyway.

src/planwise/client/projects/handlers.cljs
fgasperij: I'll add the nil case.

src/planwise/client/projects/handlers.cljs
fgasperij: Changed it for hint.

src/planwise/client/projects/handlers.cljs
```diff
@@ -240,33 +226,52 @@
...
+(defn initiate-project-update [db request-with]
+  (let [new-db
+        (case (get-in db [:map-state :current])
+          :loaded            (loaded-to-request-pending db)
+          :request-pending   (request-pending-to-request-pending db)
+          :loading           (loading-to-request-pending db)
+          :loading-displayed (loading-displayed-to-request-pending db))]
+    (-> new-db
+        (assoc-in [:map-state :current] :request-pending)
+        (assoc-in [:map-state :request-with] request-with))))
```
ggiraldez: Three small details while you rebase/merge your changes:

1. instead of `-to-` use the arrow `->`, eg. `loaded->request-pending`... but:
2. all the `-to-request-pending` functions don't actually change the state to ':request-pending'; I'd go for inlining those functions here
3. I meant the vertical layout is more readable in all cases, not just `loading-to-request-pending`

Issue #109, opened 2016-08-01
Display population colour scale reference on map


spalladino:
Show a legend with the mapping between colours and population densities

mmuller:
Ok in version v0.4.0-178-g80a5a8a

PR #175, opened 2016-08-23
Legend for map color scale

juanedi:
See #109.

### DIFF
--- a/resources/planwise/public/js/leaflet.legend.js
+++ b/resources/planwise/public/js/leaflet.legend.js
@@ -1,0 +1,39 @@
+// goog.provide("leaflet.control.legend");
+
+var formatNumber = function(n) {
+  var nKm2 = n * 100; // n is expressed in 100m2
+
+  if ((millions = nKm2 / 1000000) > 1) {
+    return Math.round(millions) + "M";
+  } else if ((thousands = nKm2 / 1000) > 1) {
+    return Math.round(thousands) + "K";
+  } else {
+    return nKm2.toString();
+  }
+}
+
+L.Control.Legend = L.Control.extend({
+  onAdd: function() {
+    var container = L.DomUtil.create('div', 'legend');
+
+    var title = L.DomUtil.create("div", "title", container);
+    title.innerHTML = "Population / Km<sup>2</sup>";
+
+    var content = L.DomUtil.create("div", "color-scale", container);
+
+    var bottom = L.DomUtil.create("span", "", content);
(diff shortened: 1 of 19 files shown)


resources/planwise/public/js/leaflet.legend.js
```diff
@@ -0,0 +1,39 @@
...
+    L.DomUtil.create("div", "", colorBarContainer);
+
+    var max = L.DomUtil.create("span", "", content);
+    max.innerText = formatNumber(this.options.legendMax);
+
+    return container;
+  }
+})
+
+L.control.legend = function(id, options) {
```
ggiraldez: You are missing at least this entry in the `externs.js` file.

resources/sass/_map.scss
```diff
@@ -1,3 +1,42 @@
...
+      position: relative;
+      top: 2px;
+
+      :first-child {
+        height: 100%;
+        width: 100%;
+        opacity: 0.5;
+
+        // this gradient should be kept in sync with mapserver's class definitions.
+        background: linear-gradient(to right, #FFC800 0%,#FF9600 8%,#FF6400 20%,#FF3200 39%,#FF1400 59%,#FF0000 100%);
```
ggiraldez: I don't like that this being data is saved as a static style in a CSS file. Can't this be saved in the database and injected via properties from the reagent component?

scripts/regions-population
```diff
@@ -8,29 +8,45 @@ DATA_PATH=${DATA_PATH:-data}
...
-      $POPULATION_FILE $TARGET
+      $POPULATION_FILE $TMP_FILE
+
+    echo " Aggregating population for ${id}"
+    IFS=' ' read -a AGGREGATES <<< $(${BIN_PATH}/aggregate-population ${TMP_FILE})
+    POPULATION_SUM=${AGGREGATES[0]}
+    POPULATION_MAX=${AGGREGATES[1]}
+
+    echo " Normalizing raster file scale for ${id}"
+    gdal_translate -ot Byte -scale 0 $POPULATION_MAX 0 255 $TMP_FILE $TARGET
```
ggiraldez: We lose precision (at the very minimum) for the demand calculations. Talking with @spalladino a good solution would be to save the original raster and use a VRT with the scaling operation for rendering with Mapserver.

spalladino: Actually, if we save both the original and the translated raster, we would not need a VRT at all. But this can all wait until we rework the demand calculation.

PR #176, opened 2016-08-24
SVG icons

spalladino:


src/planwise/client/components/common.cljs
```diff
@@ -6,6 +6,15 @@
   [:div.loading
    [:h3 "Loading..."]])
 
+;; SVG based icons
+
+(defn icon [name & [class]]
+  [:svg {:class (or class "icon")
```
ggiraldez: One small minor thing here: you can do
```clojure
(defn icon [icon-name & [class]]
  (let [icon-name (some-> icon-name name)]
    [:svg ...]))
```
to be able to use a keyword as in `(icon :car)` as well as a string `(icon "car")`

spalladino: Fixed in 9781799e316eeecc81f355f893f0a0f1d7efe04d

PR #178, opened 2016-08-25
150 improve site navigation

fgasperij:


src/planwise/client/projects/db.cljs
```diff
@@ -35,7 +35,13 @@
...
    :demand-map-key nil ;; string
    :unsatisfied-count nil ;; number
-   :project-data   {}}) ;; {:keys id goal region-id stats filters region-population region-area-km2}
+   :project-data   {} ;; {:keys id goal region-id stats filters region-population region-area-km2}
+   :wizard         {
+                    :set false
+                    :tabs {
+                      :demographics    :unvisited
+                      :facilities      :unvisited
+                      :transport       :unvisited}}})
```
spalladino: I'm not sure I like the model for the wizard. Typically wizards are modelled using the _step_ the user is currently in. So `:wizard` could be simply one of `:demographics`, `:facilities`, `:transport`, `:finished`; and then a _subscription_ could answer whether a particular tab should be displayed as visited or not, based on the current step.

src/planwise/client/components/nav.cljs
```diff
@@ -3,10 +3,12 @@
...
 ;; Navigation components
 
-(defn- li-menu-item [{:keys [selected item href title icon]}]
+(defn- li-menu-item [{:keys [selected item href title icon wizard-state]}]
   (let [is-selected? (or (= selected item)
                          (item selected))]
-    [:li {:class (when is-selected? "active")}
+    [:li {:class (str
+                  (when is-selected? "active ")
+                  (when wizard-state (name wizard-state)))}
```
spalladino: `(some-> wizard-state name)`

src/planwise/client/projects/components/sidebar.cljs
```diff
@@ -95,7 +124,9 @@
          :label-fn :name
          :on-change #(dispatch [:projects/set-transport-time %])
          :model transport-time]
-        (icon :car)]])))
+        (icon :car)]
+       (when @wizard-mode-on
+         (next-and-back-buttons :transport (:id @current-project)))])))
```
spalladino: I'd try to move all three invocations of ` (when @wizard-mode-on (next-and-back-buttons tab-name (:id @current-project)))` from each tab-filters method to `sidebar-section`, something more or less like:

```
(defn sidebar-section [selected-tab]
  [:aside
   (case selected-tab
     :demographics [demographics-filters]
     :facilities   [facility-filters]
     :transport    [transport-filters])])
   (when @wizard-mode-on
     (next-and-back-buttons selected-tab (:id current-project)))])))
```

src/planwise/client/projects/handlers.cljs
```diff
@@ -111,16 +112,39 @@
...
+(defn- update-next-tab-state [db next-tab]
+  (if (= :unvisited (get-in db [:current :wizard :tabs next-tab]))
+    (assoc-in db [:current :wizard :tabs next-tab] :visiting)
+    db))
+
+(defn- update-wizard-state [db next-tab]
+  (-> db
+      (update-next-tab-state next-tab)
+      (update-tabs-state next-tab)))
+
```
spalladino: Changing the model would yield these methods unnecessary: instead of recalculating the state of each tab when the wizard state changes, simply set the new step, and the `reaction`s in the subscriptions should re-calculate the _visited_ state for each tab as needed.

src/planwise/client/projects/handlers.cljs
```diff
@@ -111,16 +112,39 @@
...
+   (let [changed-project? (not= project-id (get-in db [:current :project-data :id]))
+         new-db (if changed-project?
+                  db
+                  (update-wizard-state db section))]
+     (if changed-project?
+       (dispatch [:projects/load-project project-id (section->with section)])
+       (case section
+         (:facilities :transport) (dispatch [:projects/load-facilities])
+         nil))
+     new-db)))
```
spalladino: It looks awkward to have two `ifs` on the same condition, right next to each other. Maybe we could merge them?

fgasperij: If we do then two `do`s should be added, I'm not sure that's less awkward.

spalladino:
Please squash "WIP", "almost done" and build-failed commits

resources/svg/README.md
```diff
@@ -0,0 +1,12 @@
...
+1- Search for them (here)[https://design.google.com/icons/]
+2- Bajalos en 24dp, negro, SVG y tiralos en esa carpeta
+2- Download them in the icons subfolder as:
+   - 24dp
+   - black
+   - SVG
+3- Editá el SVG con tu editor de texto favorito para q quede igual a los demás, ie: agregar un tag de <title>, borrar un fill q pueda haber a nivel del tag <svg>, borrar height y width hardcodeados (pero dejar viewbox)
+3- Edit them so it follows the current norm. Add it a title tag, remove the fill attribute from the svg tag, the height and the width.
+4- Run `lein build-icons`
+5- Draw it with `planwise.client.components.common/icon`.
```
ggiraldez: This should be *all* written in English.

resources/svg/README.md
```diff
@@ -0,0 +1,12 @@
+## Add new icons
+
+1- Search for them (here)[https://design.google.com/icons/]
+2- Bajalos en 24dp, negro, SVG y tiralos en esa carpeta
+2- Download them in the icons subfolder as:
+   - 24dp
+   - black
+   - SVG
+3- Editá el SVG con tu editor de texto favorito para q quede igual a los demás, ie: agregar un tag de <title>, borrar un fill q pueda haber a nivel del tag <svg>, borrar height y width hardcodeados (pero dejar viewbox)
+3- Edit them so it follows the current norm. Add it a title tag, remove the fill attribute from the svg tag, the height and the width.
```
spalladino: English please! ;-)

src/planwise/client/components/nav.cljs
```diff
@@ -3,12 +3,17 @@
...
                          (item selected))]
-    [:li {:class (when is-selected? "active")}
+    [:li {:class (str
+                  (when is-selected? "active ")
+                  (when wizard-state (str (name wizard-state) " wizard")))}
      [:a {:href href}
       (some-> icon (components/icon "icon-small"))
+      (when (and wizard-state (> tab-number 1)) [:div.separator])
+      (when (and wizard-state (= wizard-state :visited)) (components/icon :check-circle-wizard "icon-small wizard-check"))
+      (when (and wizard-state (not= wizard-state :visited)) [:div.wizard-number tab-number])
```
spalladino: I don't like coupling the `li-menu-item` function with the wizard. I'd rather have the function accept a generic content to be rendered along with or instead of the title, and work the wizard-state from outside the function. You could even have a `li-wizard-menu-item` function that calls `li-menu-item` with this.

src/planwise/client/components/nav.cljs
```diff
@@ -3,12 +3,17 @@
...
+(defn- li-menu-item [{:keys [selected item href title icon tab-number wizard-state]}]
   (let [is-selected? (or (= selected item)
                          (item selected))]
-    [:li {:class (when is-selected? "active")}
+    [:li {:class (str
+                  (when is-selected? "active ")
+                  (when wizard-state (str (name wizard-state) " wizard")))}
      [:a {:href href}
       (some-> icon (components/icon "icon-small"))
+      (when (and wizard-state (> tab-number 1)) [:div.separator])
```
spalladino: See if this can be solved using CSS via `::after` selector with `content` property, with a `:not(:first-child)` if needed.

src/planwise/client/components/nav.cljs
```diff
@@ -3,12 +3,17 @@
...
 ;; Navigation components
 
-(defn- li-menu-item [{:keys [selected item href title icon]}]
+(defn- li-menu-item [{:keys [selected item href title icon tab-number wizard-state]}]
   (let [is-selected? (or (= selected item)
                          (item selected))]
-    [:li {:class (when is-selected? "active")}
+    [:li {:class (str
+                  (when is-selected? "active ")
+                  (when wizard-state (str (name wizard-state) " wizard")))}
```
ggiraldez: The `wizard` class seems redundant here. Also, string catenation is not ideal. Use a collection (eg. set) and `clojure.string/join`.

src/planwise/client/components/nav.cljs
ggiraldez: The separator is redundant.

src/planwise/client/components/nav.cljs
ggiraldez: This should be replaced by a single `if`

src/planwise/client/current_project/handlers.cljs
```diff
@@ -48,20 +48,45 @@
...
 (defn- project-loaded [db project-data]
-  (dispatch [:current-project/fetch-facility-types])
-  (dispatch [:regions/load-regions-with-geo [(:region-id project-data)]])
-  (db/new-viewmodel project-data))
+  (let [new-db (db/new-viewmodel project-data)]
+    (dispatch [:current-project/fetch-facility-types])
+    (dispatch [:regions/load-regions-with-geo [(:region-id project-data)]])
+    (if (:is-new? project-data)
+      (assoc-in new-db [:wizard :set] true)
+      new-db)))
```
spalladino: `(assoc-in new-db [:wizard :set] (:is-new? project-data))`

src/planwise/client/current_project/components/header.cljs
```diff
@@ -5,29 +5,39 @@
...
+                    :title "Transport Means"
+                    :icon :transport-means}
+                   #_{:item :scenarios
+                      :href (routes/project-scenarios route-params)
+                      :title "Scenarios"}]]
+    (if wizard-mode-on
+      (->> tab-items
+           (mapv #(assoc %
+                         :wizard-state ((:item %) wizard-tabs-state)
+                         :tab-number (+ 1 (.indexOf [:demographics :facilities :transport] (:item %)))))
```
ggiraldez: Meet http://clojuredocs.org/clojure.core/map-indexed

src/planwise/client/current_project/components/sidebar.cljs
```diff
@@ -15,6 +17,28 @@
...
+    :transport       (routes/project-transport {:id project-id})))
+
+(defn- next-and-back-buttons [tab-name project-id]
+  (let [tabs [:demographics :facilities :transport]
+        has-next? (not= (last tabs) tab-name)
+        has-back? (not= (first tabs) tab-name)
+        next (second (drop-while #(not= tab-name %) tabs))
+        back (last (take-while #(not= tab-name %) tabs))]
+    [:div.nav-buttons {:class (if (and has-back? has-next?) "both" "just-one")}
+      (if has-back?
```
ggiraldez: `when` is more idiomatic if there is no else expression

src/planwise/client/current_project/components/sidebar.cljs
```diff
@@ -15,6 +17,28 @@
...
    [:div.stat-value value]])
 
+(defn- route-by-tab-name [tab-name project-id]
+  (case tab-name
+    :demographics    (routes/project-demographics {:id project-id})
+    :facilities      (routes/project-facilities {:id project-id})
+    :transport       (routes/project-transport {:id project-id})))
+
+(defn- next-and-back-buttons [tab-name project-id]
+  (let [tabs [:demographics :facilities :transport]
```
ggiraldez: This ordering specification is duplicated. I think we need to rethink how we handle this. Probably move all this logic into the tabs definition.

src/planwise/client/current_project/components/sidebar.cljs
```diff
@@ -15,6 +17,28 @@
...
+        has-back? (not= (first tabs) tab-name)
+        next (second (drop-while #(not= tab-name %) tabs))
+        back (last (take-while #(not= tab-name %) tabs))]
+    [:div.nav-buttons {:class (if (and has-back? has-next?) "both" "just-one")}
+      (if has-back?
+        [:div.nav-button.prev {:on-click #(accountant/navigate! (route-by-tab-name back project-id))}
+         (icon :key-arrow-left "icon-small")
+         [:span.prev-button-text "Prev"]])
+      (if has-next?
+        [:div.nav-button.next {:on-click #(accountant/navigate! (route-by-tab-name next project-id))}
```
ggiraldez: It makes more sense for these to be `a` elements with the appropriate hrefs. Failing that, a button is more semantically correct than a div.

src/planwise/client/current_project/db.cljs
```diff
@@ -73,7 +73,13 @@
                         :state nil                ; [nil :sending :sent]
                         :shares-search-string ""}
 
-   :project-data       nil})                      ; see ProjectData above
+   :project-data       nil                      ; see ProjectData above
+   :wizard         {
+                    :set false
```
ggiraldez: Rename to `:active?` or something more meaningful.

src/planwise/client/current_project/handlers.cljs
```diff
@@ -48,20 +48,45 @@
...
+
+(defn- update-tabs-state [db next-tab]
+  (let [tabs (get-in db [:wizard :tabs])
+        new-tabs (into {}
+                       (map (fn [[tab state]]
+                              (if (and (= state :visiting) (not= tab next-tab))
+                                [tab :visited]
+                                [tab state])) tabs))]
```
ggiraldez: Weird indentation. `tabs` should be in a new line.

src/planwise/client/current_project/handlers.cljs
```diff
@@ -48,20 +48,45 @@
...
+    (assoc-in db [:wizard :tabs] new-tabs)))
+
+(defn- update-next-tab-state [db next-tab]
+  (if (= :unvisited (get-in db [:wizard :tabs next-tab]))
+    (assoc-in db [:wizard :tabs next-tab] :visiting)
+    db))
+
+(defn- update-wizard-state [db next-tab]
```
ggiraldez: This should be named `visit-tab` or something like it and the logic can be simplified.

src/planwise/client/current_project/handlers.cljs
fgasperij: That would set the field with `nil`, not with false.

resources/svg/README.md
fgasperij: Done.

resources/svg/README.md
fgasperij: Done.

src/planwise/client/components/nav.cljs
fgasperij: Are you referring to the div or the class? What do you mean by redundant? It's part of the design.

ggiraldez: The word redundant is probably not the most adequate. Since it's a strictly visual element, it can be done entirely with CSS. Check https://jsfiddle.net/pxc0n8vj/ :)

If that's not enough, you'd be better off using `interpose` on the parent. See http://clojuredocs.org/clojure.core/interpose

src/planwise/client/current_project/components/header.cljs
fgasperij: Done.

fgasperij: Done.

src/planwise/client/current_project/components/sidebar.cljs
fgasperij: Done.

src/planwise/client/current_project/handlers.cljs
fgasperij: Done.

src/planwise/client/components/nav.cljs
fgasperij: Why do you think the `wizard` class is redundant? It's has styles associated with it.

ggiraldez: Those styles can be applied to each of the classes yielded by the different wizard states. Eg.

```css
.visited, .unvisited, .visiting {
  /* styles from wizard */
}
```

You can make the selector more specific if you're worried that you can later have clashes. Eg. `ul > li.visited, ul > li.unvisited, ul > li.visiting`

**If** that's not enough, adding a `wizard` class to the `ul` and using the selector `ul.wizard > li` is better than adding the `wizard` class to each `li`.

src/planwise/client/current_project/handlers.cljs
ggiraldez: `nil` and `false` are *usually* interchangeable. But why do you think your code sets the value to `false`?

fgasperij: Because that's the default value it has on the db, if it's not changes it stays the same.

ggiraldez: Right. My bad. Still, unless you are specifically testing with `false?` or something similar, `nil` and `false` are generally semantically equivalent.

Or... you can do:

```clojure
(assoc-in new-db [:wizard :set] (:is-new? project-data false))
```

if you really want a default value other than `nil`.

Issue #125, opened 2016-08-03
Support multi-tenancy


ggiraldez:
This is an umbrella ticket for multi-tenancy support in Planwise. A rough specification follows:

## Data ownership

- Projects should be owned by a single user.
- By default, users will only see and have access to their own projects.
- The system can support multiple datasets, eg. multiple sets of facilities.
- Datasets should be owned by a single user.
- New datasets are created by importing a collection from Resourcemap
- (For now) Datasets cannot be changed/updated once imported
- Datasets have an identifying name, a "source" Resourcemap collection and field mapping information; this will allow automatically sync'ing the dataset from Resourcemap in the future
- Projects use one dataset, selected at project's creation time
- Datasets cannot be deleted if they are being referenced in one or more projects

## Project sharing

- Projects can be shared with other users via a "connect link"
- Once a user opens the "connect link" she will have access to the shared project and the project will appear in her projects list
- Projects can only be shared for read-only access; only the owner is able to alter the filters/scenarios
- A user can choose to unfollow a shared project which will remove it from her list
- The owner user can list users who have access to the project and revoke access individually
- Project sharing confers read access to the associated dataset in the context of the project only (ie. the receiving user cannot create a new project with the shared project dataset)

nditada:
> (For now) Datasets cannot be changed/updated once imported
Not sure about that one. Why is that? If you also can't replace the dataset for a project, how do you have the facility list?

>The owner user can list users who have access to the project and revoke access 
That's a NTH, we can leave that in the backlog for the end of the project.

>Project sharing confers read access to the associated dataset in the context of the project only 
I'm indifferent to that. We can go either way, depending on what's easiest to implement (i.e. it's ok if you can use the shared facility list in another project, I see no problem).

ggiraldez:
>> (For now) Datasets cannot be changed/updated once imported

> Not sure about that one. Why is that? If you also can't replace the dataset for a project, how do you have the facility list?

This is just to make the initial implementation simple, but can be added in the next release (or this one should time allow it). Since the dataset will keep the source collection from Resourcemap as well as the mappings, it will be possible to update the facility list by importing the same collection again.

>> Project sharing confers read access to the associated dataset in the context of the project only

> I'm indifferent to that. We can go either way, depending on what's easiest to implement (i.e. it's ok if you can use the shared facility list in another project, I see no problem).

[...]

PR #179, opened 2016-08-26
Feature/datasets ownership

### LINKED ISSUE #154: Support multiple datasets
- Add dataset ownership
- Add dataset reference to projects
- Redo datasets page according to mockups
- Change importer to support job queueing

See #125 for more details.

ggiraldez:
Closes #154 and #125 

- Add ownership to datasets
- Projects now require a dataset to be created and will use the facilities from that dataset only
- Creating a dataset starts the import process
- Multiple datasets can be created simultaneously


src/planwise/client/datasets/components/listing.cljs
```diff
@@ -0,0 +1,74 @@
...
+          :disabled cancelling?}
+         (if cancelling? "Cancelling..." "Cancel")]]
+
+       (zero? project-count)
+       [:div.bottom-right
+        [:button.delete
+         {:on-click (utils/with-confirm
+                      #(dispatch [:datasets/delete-dataset id])
+                      "Are you sure you want to delete this dataset?")}
+         "\u2716 Delete"]])]))
```
spalladino: We might want to add a "Cannot delete as there are projects associated" label, to explain to the user why a dataset can be deleted while another one is not offering that option

src/planwise/client/datasets/handlers.cljs
```diff
@@ -1,139 +1,172 @@
...
-         (assoc-in [:resourcemap :authorised?] (:authorised? datasets-info))
-         (assoc-in [:resourcemap :collections] (:collections datasets-info))
-         (assoc :state (db/server-status->state server-status)
-                :server-status server-status
-                :facility-count (:facility-count datasets-info))))))
+ (fn [db [_ time]]
+   (let [sets (asdf/value (:list db))
+         statuses (map :server-status sets)
+         any-running? (some some? statuses)
+         refresh-interval (if any-running? 1000 10000)
```
spalladino: Why do we need to keep refreshing if there is none updating?

PR #183, opened 2016-08-26
Display importer warnings in datasets

### LINKED ISSUE #89: Display a message notifying users when the application doesn’t have routing information
We can display an alert when a user is importing sites outside the area where we have routing information.

We could display something like "warning: there are some sites outside the area where we have routing information"

spalladino:
Fixes #89 

screen shot 2016-08-26 at 7 08 18 pm

### DIFF
--- a/src/planwise/component/importer.clj
+++ b/src/planwise/component/importer.clj
@@ -43,10 +43,10 @@
     {:code (:code type-field)
      :options new-options}))
 
-(defn sites-with-location
-  "Filter Resourcemap sites which have a valid location"
-  [sites]
-  (filter #(and (:lat %) (:long %)) sites))
+(defn with-location
+  "Filter function for valid location"
+  [{:keys [lat long], :as site}]
+  (and lat long))
 
 (defn facility-type-ctor
   "Returns a function which applied to a Resourcemap site returns the facility
@@ -75,7 +75,7 @@
   using the given facility type field definition."
   [sites type-field]
   (->> sites
-       (sites-with-location)
+       (filter with-location)
        (map (site->facility-ctor type-field))))
 
 (defn import-collection-page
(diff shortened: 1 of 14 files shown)


src/planwise/component/importer.clj
```diff
@@ -89,18 +89,23 @@
...
 (defn process-facilities
   [facilities facility-ids]
-  (doseq [id facility-ids]
-    (info "Processing facility" id)
-    (facilities/preprocess-isochrones facilities id)))
+  (map
+    (fn [id]
+      (info "Processing facility" id)
+      (facilities/preprocess-isochrones facilities id))
+    facility-ids))
```
ggiraldez: `map` returns a lazy sequence which may not be realised in the right moment, or the correct thread for that matter. Use `doall`, `mapv` or `(into [] (map ..))` to force immediate computation.

src/planwise/model/import_job.clj
```diff
@@ -142,15 +143,26 @@
         (assoc :result :update-projects-failed
                :error-info error))))
 
+(defn- update-facilities-counts [job event]
+  (let [[_ _ results] event
+        no-road-network (filter #{:no-road-network} results)
+        outside-regions (filter #{:outside-regions} results)]
+    (-> job
+      (update :facilities-without-road-network-count + (count no-road-network))
+      (update :facilities-outside-regions-count + (count outside-regions)))))
```
ggiraldez: Meet http://clojuredocs.org/clojure.core/frequencies :)

spalladino: Fixed in 35598b6f50a5ae4d1229d19bbf3f9086c5b92cea

src/planwise/component/importer.clj
spalladino: Fixed in b804d2bf53ec3a24300f80b5e62e5170eeb49c8b

PR #192, opened 2016-08-30
Refactor client's project viewmodel

### LINKED ISSUE #121: Split projects list and current project view model in the client
Everything is under a current namespace and database key. Split into `:projects` and `:current-project` and the corresponding namespaces.

ggiraldez:
Closes #121 

- Splits projects view model into listing and current project view model
- Separate code to handle projects list and currently active project into different namespaces
- Simplify current project view model in the client DB, de-duplicating many filters and other pieces of information; keep everything not view state related or associations inside the `:project-data` key

### DIFF
--- a/src/planwise/client/current_project/subs.cljs
+++ b/src/planwise/client/current_project/subs.cljs
@@ -27,13 +27,12 @@
 (register-sub
  :current-project/facilities
  (fn [db [_ data]]
-   (let [facility-data (reaction (get-in @db [:current-project :facilities]))]
-     (reaction
-      (case data
-        :filters (:filters @facility-data)
-        :isochrones (:isochrones @facility-data)
-        :filter-stats (select-keys @facility-data [:count :total])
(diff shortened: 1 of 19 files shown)


src/planwise/client/current_project/subs.cljs
```diff
@@ -60,12 +59,12 @@
...
  :current-project/transport-time
  (fn [db [_]]
-   (reaction (get-in @db [:current-project :transport :time]))))
+   (reaction (get-in @db [:current-project :project-data :filters :transport :time]))))
 
 (register-sub
  :current-project/demand-map-key
  (fn [db [_]]
-   (reaction (get-in @db [:current-project :demand-map-key]))))
+   (reaction (get-in @db [:current-project :project-data :demand-map-key]))))
```
spalladino: Though not being used currently, `demand-map-key` is not actually an attribute of the project per se, but of its region+filters. Should we still keep it in `project-data`?

PR #200, opened 2016-09-01
Project sharing

### LINKED ISSUE #155: Project sharing
- Allow users to share projects by generating and sharing a link
- Opening the link gives read access to the project to anyone
- List shared projects for a user
- Allow opening a shared project in read-only mode

See #125.

### LINKED ISSUE #156: Project sharing via email
When sharing a project, allow the user to enter a list of mail addresses to send the sharing link to.

Depends on #155.
See also #125.

spalladino:
Add a sharing dialog to projects, where the owner can:
- Copy the share link
- Reset the share link
- Review which users have access to the project
- Revoke access for any of them
- Send the share link via email

Any user can follow the share link to have the project added to his/her list of projects, in read-only mode, with an option to leave the shared project. Shared projects are listed in the index with the email of the project owner who shared it.

Fixes #155 
Fixes #156

screen shot 2016-09-01 at 6 07 41 pm


src/planwise/client/current_project/components/header.cljs
```diff
@@ -23,15 +23,30 @@
...
+                    "Are you sure you want to leave this shared project?")}
+        (common/icon :exit "icon-small")
+        "Leave project"]]
+      [:div
+       [:button.secondary
+        {:on-click (utils/prevent-default
+                     #(dispatch [:current-project/open-share-dialog]))}
+        (common/icon :share "icon-small")
+        (if (pos? share-count)
+          (str "Shared with " share-count)
```
ggiraldez: Add the "user(s)" string at the end of the legend. Ie.

```clojure
(str "Shared with " (utils/pluralize share-count "user"))
```

src/planwise/client/current_project/components/sharing.cljs
```diff
@@ -0,0 +1,88 @@
...
+    (fn []
+      (let [close-fn #(dispatch [:current-project/close-share-dialog])
+            key-handler-fn #(when (= 27 (.-which %)) (close-fn))
+            select-text-fn #(.select (.-target %))
+            reset-share-token-fn (utils/with-confirm
+                                   #(dispatch [:current-project/reset-share-token])
+                                   (str "Are you sure you want to reset the sharing link?\n\n"
+                                         "Anyone who has not yet opened the link will not "
+                                         "be able to access, and will need to be shared the new "
+                                         "link."))]
```
ggiraldez: The wording here is a bit weird. How about "If you reset the link, anyone who has not yet accessed the project will not be able to do it until you invite them again."

src/planwise/client/current_project/handlers.cljs
```diff
@@ -65,6 +72,34 @@
...
+     ; 1- User navigates to /project/:id/access/:token
+     ; 2- Requests for projects list and request access to shared project are sent to the server
+     ; 3- The server processes the projects list before the access request, so the list does not contain the new project
+     ; 4- The server processes the request access and sends the response to the client before the response for (3)
+     ; 5- The client optimistically adds the new project to the list and asdf/invalidates it
+     ; 6- The response for (3) arrives at the client, which updates the list *without* the shared project, and marks the asdf as valid
+     ; This should be easily fixed when asdf supports versioning of data, then
+     ; the load-projects dispatch can be replaced with:
+     ; (dispatch [:projects/invalidate-projects [project-data]])
+     (dispatch [:projects/load-projects])
```
ggiraldez: I'm not convinced this is not the same as simply invalidating the project list (without optimistic updates). In that case eventually the view will force a `load-projects` anyway. Am I missing something?

src/planwise/client/current_project/handlers.cljs
```diff
@@ -219,3 +262,101 @@
...
+;; ---------------------------------------------------------------------------
+;; Sharing handlers
+
+(register-handler
+ :current-project/open-share-dialog
+ in-current-project
+ (fn [db [_]]
+   ; TODO: Fire a request to update the project shares when opening the dialog to ensure it is up to date
+   ; What happens if the user deletes an item from the list before the request returns? In that case,
+   ; the item that was just deleted will re-appear.
```
ggiraldez: In the view, you can decide to render the share list only if the data is valid by using `asdf/valid?`. It's not the best user experience, but it eliminates this particular concern.

src/planwise/client/current_project/db.cljs
```diff
@@ -58,21 +65,33 @@
                         :request nil
                         :request-with nil}
 
+   :view-state         :project                   ; [:project :share-dialog]
+
+   :shares             (asdf/new [])              ; [ProjectShare]
+   :sharing            {:emails-text nil
+                        :token-state nil          ; [nil :reloading]
```
ggiraldez: Why not use `asdf` here? It does require `dissoc`ing the token before updating `project-data` (to avoid duplication), but since it's a special piece of info that can be updated individually it makes sense. Otherwise, use an `asdf` inside `project-data`. In that case updates to `project-data` need to be controlled but I think it needs to change in `db/update-viewmodel` only.

src/planwise/client/current_project/handlers.cljs
```diff
@@ -219,3 +262,101 @@
...
+ :current-project/share-token-loaded
+ in-current-project
+ (fn [db [_ {token :token}]]
+   (-> db
+     (assoc-in [:project-data :share-token] token)
+     (assoc-in [:sharing :token-state] nil))))
+
+(defn- remove-project-share
+  [coll user-id]
+  (remove #(= user-id (:user-id %)) coll))
```
ggiraldez: Obsolete left over :)

src/planwise/client/current_project/handlers.cljs
```diff
@@ -219,3 +262,101 @@
...
+   ; If we do so, we risk loading a stale list if:
+   ; 1- The user an item from the list
+   ; 2- We optimistically remove it and issue the delete request
+   ; 3- The request returns from the server and we invalidate the list
+   ; 4- The invalidate triggers a full reload of the list
+   ; 5- The user deletes another item, which is optimistically deleted
+   ; 6- The full reload of the list resolves, adding the item just deleted by the user
+   ; Should we cancel the previous request to reload the shares list?
+   ; Or simply ignore the invalid status of the list?
+   (update db :shares asdf/invalidate! remove-by :user-id user-id)))
```
ggiraldez: I don't think these corner cases can be fixed without versioning. And even with versioning if you do optimistic updates to the db you may still have transient inconsistent UI states. We need to think this further.

src/planwise/client/current_project/subs.cljs
```diff
@@ -96,3 +135,47 @@
...
+                                              (string/join ", " invalid-emails) " "
+                                              (if (= 1 (count invalid-emails))
+                                                "is an invalid email address"
+                                                "are invalid email addresses"))
+                      :else                "Send this project sharing link to the emails above")
+          :label    (cond
+                      (= :sending state) "Sending..."
+                      (= :sent state)    "Sent"
+                      (seq valid-emails) (str "Send to " (utils/pluralize (count valid-emails) "user"))
+                      :else              "Send")})))))
```
ggiraldez: There is some view-only related code here that I think belongs in a Reagent component. Eg. title and label, probably editable and disabled as well. Otherwise the `share-dialog` component is tightly coupled to this subscription for no particular reason.

src/planwise/client/mapping.cljs
```diff
@@ -45,7 +45,7 @@
 ;; only "classic" mapbox styles can be used with this static API.
 ;; the new API does not allow defining an overlay as a query parameter.
 (defn static-image [geojson]
-  (fmt/format "https://api.mapbox.com/v4/%s/geojson(%s)/auto/256x144.png?access_token=%s"
+  (fmt/format "https://api.mapbox.com/v4/%s/geojson(%s)/auto/256x158.png?access_token=%s"
```
ggiraldez: Is there a nice way of decoupling this request from the CSS file?

src/planwise/client/projects/handlers.cljs
```diff
@@ -22,8 +22,10 @@
 (register-handler
  :projects/invalidate-projects
  in-projects
- (fn [db [_]]
-   (update db :list asdf/invalidate!)))
+ (fn [db [_ new-projects]]
+   (if new-projects
+     (update db :list asdf/invalidate! into new-projects)
```
ggiraldez: This branch is not used. And I don't really like it. I'd rather invalidate-optimistically-update in-place.

src/planwise/client/utils.cljs
```diff
@@ -55,6 +73,18 @@
...
      (gstring/format format-string percentage))))
 
+;; Collection manipulation
+
 (defn remove-by-id
   [coll id]
   (remove #(= id (:id %)) coll))
+
+(defn remove-by
+  [coll field value]
```
ggiraldez: Nice! `field` can be an arbitrary function though.

src/planwise/endpoint/projects.clj
```diff
@@ -12,25 +12,28 @@
...
+                                time       (get-in project [:filters :transport :time])
+                                isochrones (when time
+                                             []) ; TODO: Load isochrones to calculate demand
+                                demand     (when isochrones
+                                             (maps/demand-map maps (:region-id project) isochrones))]
+                            (-> project
+                              (merge demand)
+                              (assoc :facilities project-facilities)))
+                          project)]
+    (if (:read-only project)
```
ggiraldez: This should be `with-extra-data` instead of `project` right?

src/planwise/client/current_project/components/header.cljs
spalladino: Fixed in 019cede9fe9b9bc636ccd26edfc0ff9679ea4c58

src/planwise/client/current_project/components/sharing.cljs
spalladino: Fixed in 019cede9fe9b9bc636ccd26edfc0ff9679ea4c58

New text is:
> If you reset the link, anyone who has not yet accessed the project will not be able to do it until you send them the new link.

src/planwise/client/current_project/handlers.cljs
spalladino: Nope, not missing anything. The race condition described there could never happen since the handler is executed *after* the server has acknowledged the share.

Fixed in 2c5ff2d83d2a74f3530b8c61adcfcf84bbe924b8

src/planwise/client/current_project/handlers.cljs
spalladino: How about rendering the list with the invalid data, but allowing changes (ie revoking access) only after its state is `asdf/valid?` again?

src/planwise/client/current_project/db.cljs
spalladino: True. Fixed in 85ce7f3fa47247bedf665518be1fbae7435ec844

src/planwise/client/current_project/handlers.cljs
spalladino: Fixed in 62e0fab835fd4dba932666c793a17eddb996510a

src/planwise/client/mapping.cljs
spalladino: If we remove this value from the CSS, then the project cards change size when the image is actually loaded. 

In order to de-duplicate the value, I looked for some way to inject a config value into SASS as a variable but found nothing. Another option would be to keep the definition of a `$map-preview-image-size` variable in an `scss` file, and load it and parse it from clojure code to extract the value, and pass it onto the client via `config`; but it feels overkill.

src/planwise/client/projects/handlers.cljs
spalladino: Actually it is now used in `current_project/handlers` since e0027ef8b552bafa9542c26480e2f89341b73f4a. 

Though I'm not a fan of this branch either, doing the invalidate-optimistically-update in-place would imply directly modifying the projects list from the `current_project` handlers, this is, outside the `in-current-project` path.

src/planwise/endpoint/projects.clj
spalladino: @%$#^(*$*^%#^%$

Good catch. Fixed in 1df4bf00b8828da097a60b6bbe73085653c1c947.

src/planwise/client/current_project/handlers.cljs
ggiraldez: I like it!

src/planwise/client/mapping.cljs
ggiraldez: I just realized we can always use the `style` property and set the card `height` from the Reagent component. That way the height is a namespace `def` which can be used to create this URL and set the style. I think that might mess up the layout, but, it should be fixable using flex.

src/planwise/client/asdf.cljs
```diff
@@ -60,3 +60,8 @@
 (defn reload!
   [asdf]
   (assoc asdf :state :reloading))
+
+(defn update
+  "Updates the value in the asdf preserving its current state"
+  [asdf update-fn & args]
+  (assoc asdf :value (apply update-fn (:value asdf) args)))
```
ggiraldez: Erm... this is exactly what `swap!` does :)

src/planwise/client/current_project/subs.cljs
```diff
@@ -20,6 +25,42 @@
...
+ :current-project/share-link
+ (fn [db [_]]
+   (reaction
+     (let [token (get-in @db [:current-project :sharing :token])
+           project-id (get-in @db [:current-project :project-data :id])]
+       (asdf/update token (fn [t]
+                            (str
+                              (.-origin js/document.location)
+                              (routes/project-access {:id project-id
+                                                      :token t}))))))))
```
ggiraldez: Why would you update the token value here? This could lead to an invalid link if the reaction runs twice. Why not simply return the link as the value of the reaction?

src/planwise/client/asdf.cljs
ggiraldez: Well, not exactly since this doesn't reset the state. But this shouldn't be needed. See my other comment.

src/planwise/client/current_project/subs.cljs
spalladino: I want this reaction to return an `asdf` with:
- state: same as the token `asdf`
- value: the link created from the token

So the update is used to return a new asdf from the token's one, that preserves the state and modifies the value to return a link. Note that `db` is not modified anywhere, so the reaction could be run any number of times without issues, as the token's asdf is not changed (yay immutability!).

src/planwise/client/mapping.cljs
spalladino: Good idea. Fixed in 0e687f2f47ebed9bf2d3c6838cc370c532c8d4a2.

src/planwise/client/current_project/subs.cljs
ggiraldez: Doh! Of course, this is a subscription. Sorry for the confusion.

src/planwise/client/current_project/handlers.cljs
spalladino: Implemented in 1dd61096d1eacdef4db4a155fd519a713f98291f

src/planwise/client/current_project/handlers.cljs
spalladino: Agreed. Removed this in 1dd61096d1eacdef4db4a155fd519a713f98291f, in favour of just reloading the list of shares when opening the share dialog.

src/planwise/client/current_project/subs.cljs
spalladino: True. Removed this logic from the subscription, and split the share-dialog into components in 765e7970b955bea3c0ab5cdf90905723db539d98.

PR #206, opened 2016-09-06
Show dataset warnings

### LINKED ISSUE #182: Display alert in project view if dataset is importing
When viewing a project whose dataset is in an importing state, display an alert banner notifying the user that some facilities may be missing isochrones, and data is still updating.

Query the server every X seconds to check if the import has finished to remove the banner.

### LINKED ISSUE #195: Show that the import process was cancelled
After cancelling an import process, there's no indication in the dataset that the import didn't complete because it was cancelled by the user.

Import result is being saved to the database though (as the `:result` key), just not shown in the UI.

spalladino:
Display warnings when a dataset is importing, failed, cancelled, or has warnings. This information is displayed on the dataset list, on the new project dialog, and on the project view.

Fixes #182 
Fixes #195 

screen shot 2016-09-06 at 3 37 06 pm
screen shot 2016-09-06 at 3 37 17 pm
screen shot 2016-09-06 at 3 37 28 pm

### DIFF
--- a/src/planwise/client/datasets/db.cljs
+++ b/src/planwise/client/datasets/db.cljs
@@ -126,3 +126,17 @@
   [new-dataset-data]
   (and (some? (:collection new-dataset-data))
        (some? (:type-field new-dataset-data))))
+
+(defn dataset->status
+  "Returns one of importing, cancelled, success or unknown, based on the dataset server status and import result"
+  [{:keys [server-status import-result], :as dataset}]
+  (when dataset
+    (case (keyword (:status server-status))
+      :importing    :importing
+      :cancelling   :cancelled
+      :unknown      :unknown
+
+      (case (keyword (:result import-result))
+        :cancelled    :cancelled
+        :success      :success
+        :error))))
--- a/src/planwise/client/datasets/db.cljs
+++ b/src/planwise/client/datasets/db.cljs
@@ -83,29 +83,6 @@
   (or (= :cancel-requested state)
       (= :import-requested state)))
 
-(defn server-status->string
-  [server-status]
-  (case (:status server-status)
-    (nil :ready :done)
-    "Ready to use"
-
-    :importing
-    (let [progress (:progress server-status)
-          progress (when progress (str " " (format-percentage progress)))]
-      (case (:state server-status)
-        :start "Waiting to start"
-        :importing-types "Importing facility types"
-        (:request-sites :importing-sites) (str "Importing sites from Resourcemap" progress)
-        (:processing-facilities) (str "Pre-processing facilities" progress)
-        (:update-projects :updating-projects) "Updating projects"
-        "Importing..."))
-
-    :cancelling
-    "Cancelling..."
-
-    :unknown
-    "Unknown server status"))
-
 (defn import-progress
   [server-status]
   (case (:status server-status)
@@ -128,17 +105,6 @@
         1))
 
     0))
-
-(defn import-result->string
-  [result]
-  (case result
-    :success "Success"
-    :cancelled "Cancelled"
-    :unexpected-event "Fatal error: unexpected event received"
-    :import-types-failed "Error: failed to import facility types"
-    :import-sites-failed "Error: failed to import sites from Resourcemap"
-    :update-projects-failed "Error: failed to update projects"
-    nil))
 
 (defn server-status->state
   [{status :status}]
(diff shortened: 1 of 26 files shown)


src/planwise/client/datasets/db.cljs
```diff
@@ -160,3 +126,54 @@
...
+
+(defn dataset->status-class
+  "Returns a status class relative to the dataset status"
+  [dataset]
+  (case (dataset->status dataset)
+    (nil :success :importing) nil
+    (:error :unknown)         "error"
+    :warn                     "warning"
+    :cancelled                "cancelled"
+    nil))
```
ggiraldez: Why move `import-result->string` and `dataset-status->string` into the view namespaces but not these three?

src/planwise/client/projects/components/new_project.cljs
```diff
@@ -56,7 +59,11 @@
            :label-fn :name
            :filter-box? true
            :on-change #(reset! new-project-dataset-id %)
-           :model new-project-dataset-id]]
+           :model new-project-dataset-id]
+          (when-let [warning (dataset->warning-text @new-project-dataset)]
+           [:p.notice.dataset-status
+            {:class (dataset->status-class @new-project-dataset)}
+            warning (dataset->status-class @new-project-dataset)])]
```
ggiraldez: Why is the class rendered as content?

src/planwise/client/datasets/db.cljs
spalladino: Because they are re-used from the projects view namespaces. However, now that I think of it, there is nothing wrong in a project view depending on a component in the dataset namespace, something like `client/datasets/components/status.cljs`. Should I move them there?

src/planwise/client/utils.cljs
```diff
@@ -73,7 +73,11 @@
          format-string (str "%." decimals "f%%")]
      (gstring/format format-string percentage))))
 
-;; Collection manipulation
+;; Collection utils
+
+(defn find-by-id
+  [coll id]
+  (first (filter #(= id (:id %)) coll)))
```
ggiraldez:
```clojure
(some #(when (= id (:id %)) %) coll)
```

should be more efficient. Or even:

```clojure
(reduce #(when (= id (:id %2)) (reduced %2)) nil coll)
```

Ref: http://stackoverflow.com/a/32405094/718209

src/planwise/client/datasets/db.cljs
```diff
@@ -160,3 +126,54 @@
...
+          :error)))))
+
+(defn dataset->warning-text
+  "Returns a warning text if the dataset status is importing, cancelled, unknown or error"
+  [dataset]
+  (case (dataset->status dataset)
+    :importing "This dataset is still being imported. Your project's data may be incomplete or inconsistent until the process finishes."
+    :cancelled "The import process for this dataset was cancelled. Your project's data may be incomplete or inconsistent."
+    :unknown   "The status for this dataset is unknown. Your project's data may be incomplete or inconsistent."
+    :error     "The import process for this dataset has failed. Your project's data may be incomplete or inconsistent."
```
ggiraldez: The wording on these messages is not clear when used in the project view. They refer to the dataset directly, but you're actually looking at a project.

src/planwise/client/datasets/db.cljs
ggiraldez: Yep, sounds good.

See my other comment regarding wording in project views. It may make sense to split some of the functions in the two namespaces.

src/planwise/client/projects/components/new_project.cljs
spalladino: Because debugging. Fixed in 1d549a2.

src/planwise/client/utils.cljs
spalladino: Fixed in d4f683f

src/planwise/client/datasets/db.cljs
spalladino: Agree. Fixed in 55f2731a3eb4071234021654613fac37bf685cee.

src/planwise/client/datasets/db.cljs
spalladino: Fixed in 77332f8aa7ddecf6b2e0512d57230be0895f19c7

Issue #191, opened 2016-08-30
Do not import sites with unknown type


juanedi:
If a resmap dataset includes site with an empty type field, we are currently creating a facility with no type. This doesn't seem right, since there is no way to visualise those facilities in the application.

We should probably ignore those sites and add a warning saying that they weren't imported because of not having a known type.

Also, consider adding a DB contraint to avoid having null references for the facility>type FK.

mmuller:
Ok in version 0.5.1-130-gab8d3ae

PR #211, opened 2016-09-07
Don't import facilities without type in resmap

juanedi:
See #191.

### DIFF
--- a/src/planwise/model/import_job.clj
+++ b/src/planwise/model/import_job.clj
@@ -101,7 +101,7 @@
   (let [page (:page job)
         facility-ids (:facility-ids job)
         facility-count (count facility-ids)
-        [_ _ [_ page-ids total-pages sites-without-location sites-without-type]] event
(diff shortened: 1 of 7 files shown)


src/planwise/model/import_job.clj
```diff
@@ -101,15 +101,16 @@
   (let [page (:page job)
         facility-ids (:facility-ids job)
         facility-count (count facility-ids)
-        [_ _ [_ page-ids total-pages sites-without-location]] event
+        [_ _ [_ page-ids total-pages sites-without-location sites-without-type]] event
```
ggiraldez: You also need to update the FSM definition, otherwise it won't import more than one page of sites from Resourcemap.

See https://github.com/instedd/planwise/commit/38fc76089601db1a93564ccd068c9dc56960476d for the last fix for a similar bug.

Also, it makes sense to re-group the return values into a hashmap. The length of that vector is beginning to get ridiculous.

PR #213, opened 2016-09-07
Re-enable fastcgi support in mapserver

### LINKED ISSUE #104: Activate FastCGI with Mapserver
Investigate why it was returning an empty tile when used for the demographics layer.

spalladino:
Fixes #104


mapserver/docker-compose.yml
```diff
@@ -11,7 +11,6 @@ mapserver:
   image: instedd/planwise-mapserver:kenya-mapserver
   volumes:
     - "../data:/data:ro"
-    - "./mapserver.conf:/etc/apache2/conf-enabled/mapserver.conf:ro"
     - "./kenya.map:/etc/mapserver/kenya.map:ro"
```
ggiraldez: The default mapfile for the image is located at `/etc/mapserver/mapserver.map` but we adding a `/etc/mapserver/kenya.map`. Since we're no longer overriding the Apache configuration, by default the CGI script will try to load the map file from a non-existent file. We should either change the Apache config or link `kenya.map` to `mapserver.map`.

I understand that going through mapcache should override this value, but still... Especially for development, if this means one less WTF moment, I consider the extra hassle worth it.

PR #214, opened 2016-09-07
Gracefully handle OAuth2 invalid grant responses.

ggiraldez:
This can happen if the user invalidated the token/authorization in
Guisso after initially authorizing Planwise. Return a nil token in this
case signalling that the user must authorize the application again.

### DIFF
--- a/src/planwise/component/auth.clj
+++ b/src/planwise/component/auth.clj
@@ -7,6 +7,7 @@
             [clojure.string :as str]
             [buddy.sign.jwt :as jwt]
             [oauthentic.core :as oauth]
+            [slingshot.slingshot :refer [try+ throw+]]
             [planwise.util.ring :refer [absolute-url]]
             [planwise.model.ident :as ident]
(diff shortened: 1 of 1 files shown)


src/planwise/component/auth.clj
```diff
@@ -140,24 +141,34 @@
...
+  and catching possible invalid grant errors and returning nil in such cases."
+  [service params]
+  (try+
+   (let [url (guisso-url service "/oauth2/token")
+         params (assoc params
+                       :client-id (:guisso-client-id service)
+                       :client-secret (:guisso-client-secret service))]
+     (-> (oauth/fetch-token url params)
+         (guisso-response->token-info)))
+   (catch [:status 400] {body :body}
```
spalladino: Who is throwing+ the whole response for this clause to catch?

ggiraldez: That would be `clj-http` used by `oauthentic`. See https://github.com/dakrone/clj-http#exceptions

Issue #199, opened 2016-08-31
Review population density scale and colors


ggiraldez:
The legend control assumes that each pixel is 100m x 100m and uses the maximum value from the raster file to display the population density. From recent measurements, the pixel size for the Kenya raster file is about 92m x 92m and may vary from country to country, depending on the data from Worldpop.

See also #142 

Since we're now using a mostly linear scale, low density areas appear empty when rendering the map. We might want to adjust the scale and use a saturated logarithmic scale, as in http://sedac.ciesin.columbia.edu/downloads/maps/grump-v1/grump-v1-population-density/kendens.jpg 

To do that, we need to review the normalization process **and** the colors used in the Mapserver configuration file. See https://drive.google.com/open?id=10qTOKQHq96aIEJ2UE0iiJSgf5jmWBBZ8qAgH1JOfFHc

Discuss with @nditada, @jkicillof and @juanedi 

![kendens](https://cloud.githubusercontent.com/assets/733591/18150153/188e4e8e-6fbb-11e6-9501-84a8c6e23a80.jpg)

nditada:
Let's make the scale look like this:
![](https://files.slack.com/files-pri/T02JWDRK4-F28Q5FDU4/screen_shot_2016-09-01_at_11.07.05.png)

juanedi:
Note that the problem regarding the pixel area size is fixed in (currently pending) PR #202.

mmuller:
Ok in version 0.5.1-130-gab8d3ae

PR #215, opened 2016-09-08
Changes in map color scales

juanedi:
See #199.

We are now saturating population rasters in 1001 people / km2. The color categories in mapserver were also changed and the map now shows a discrete legend, which shows the values when hovered.

screen shot 2016-09-08 at 7 25 19 pm

screen shot 2016-09-08 at 7 25 31 pm

### DIFF
--- a/mapserver/planwise.map
+++ b/mapserver/planwise.map
@@ -133,7 +133,7 @@ MAP
 
     CLASS
       NAME "c5"
-      EXPRESSION ([pixel] >= 255)
+      EXPRESSION ([pixel] >= 254)
       STYLE
         COLOR 255 8 0
       END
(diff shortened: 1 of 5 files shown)


mapserver/planwise.map
```diff
@@ -100,48 +100,40 @@ MAP
...
       STYLE
         COLOR 255 133 0
       END
     END
 
     CLASS
-      NAME "geq150"
-      EXPRESSION ([pixel] >= 150)
+      NAME "c5"
+      EXPRESSION ([pixel] >= 255)
```
ggiraldez: You're excluding the value `254` from the classes. This should be `[pixel] >= 254` or the previous rule should be `[pixel] < 255`.

PR #220, opened 2016-09-12
Feature/improve logging

### LINKED ISSUE #207: Review log settings for production
Also, make it that the log level is at least configurable by an environment variable. Refer to timbre documentation here: https://github.com/ptaoussanis/timbre#log-levels-and-ns-filters . It already supports environment configuration, but we are hard-coding the level in `main.clj` and `dev.clj`.

ggiraldez:
Closes #207 and adds some other logging goodies.

### DIFF
--- a/dev/dev.clj
+++ b/dev/dev.clj
@@ -21,8 +21,10 @@
             [planwise.system :as system]))
 
 ;; Logging configuration for development
-(timbre/merge-config! {:ns-blacklist ["com.zaxxer.hikari.*"]})
-(timbre/set-level! :info)
+(timbre/merge-config! {:level :debug
+                       :ns-blacklist ["com.zaxxer.hikari.*"
+                                      "org.apache.http.*"
+                                      "org.eclipse.jetty.*"]})
 
 ;; Fix JWE secret in development to facilitate debugging
 (def jwe-secret
--- a/src/planwise/main.clj
+++ b/src/planwise/main.clj
@@ -13,7 +13,7 @@
 
 (def prod-config
   {:app {:middleware     [[wrap-hide-errors :internal-error]]
-         :internal-error (io/resource "errors/500.html")}})
+         :internal-error (io/resource "planwise/errors/500.html")}})
 
 (def config
   (meta-merge config/defaults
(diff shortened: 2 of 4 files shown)


dev/dev.clj
```diff
@@ -21,8 +21,10 @@
             [planwise.system :as system]))
 
 ;; Logging configuration for development
-(timbre/merge-config! {:ns-blacklist ["com.zaxxer.hikari.*"]})
-(timbre/set-level! :info)
+(timbre/merge-config! {:level :debug
+                       :ns-blacklist ["com.zaxxer.hikari.*"
+                                      "org.apache.http.*"
+                                      "org.eclipse.jetty.*"]})
```
spalladino: Do we need to repeat the blacklist here, if it's already set in `main`?

src/planwise/main.clj
```diff
@@ -7,21 +7,28 @@
...
+         :internal-error (io/resource "planwise/errors/500.html")}})
 
 (def config
   (meta-merge config/defaults
               config/environ
               prod-config))
 
 (defn -main [& args]
-  (timbre/set-level! :warn)
+  ;; Logging configuration for development
```
spalladino: Copy-pasted comment 😜

src/planwise/util/ring.clj
```diff
@@ -34,3 +37,43 @@
...
+         exclude-uris (:exclude-uris options)]
+     (fn [request]
+       (let [uri (:uri request "")
+             params (:params request)]
+         (when-not (and (some? exclude-uris)
+                        (re-matches exclude-uris uri))
+           (log level "Starting request:"
+                (-> (:request-method request) name .toUpperCase)
+                (:uri request)
+                (or (:query-string request) ""))
```
spalladino: Maybe refactor these 3 lines into an auxiliary function, since they are repeated exactly in the function below?

dev/dev.clj
ggiraldez: `main` is not loaded during development. It's the entry point for production only.

src/planwise/main.clj
ggiraldez: Doh!

dev/dev.clj
spalladino: Good point, my bad

src/planwise/util/ring.clj
ggiraldez: Done in 3954062f9f539eba2b9910d7044601f3d88c5346

PR #222, opened 2016-09-13
Calculate unsatisfied demand on transport tab

### LINKED ISSUE #187: Calculate unsatisfied demand given the selected facilities
The algorithms to perform this calculations have been implemented already but they are not active by default. They involve precalculating catchment rasters for each facility.

Review them and fix anything we broke since they were implemented. Also review and try to reduce the required storage.

spalladino:
Fixes #187

Makes use of `calculate-demand` to calculate the raster on the fly and show it on mapserver. Also displays the satisfied demand as a progress bar on the transport tab.

Requires all facilities to be re-processed, and `calculate-demand` and `raster-isochrones` env vars to be set to "true".

screen shot 2016-09-13 at 10 06 44 am

### DIFF
--- a/src/planwise/component/maps.clj
+++ b/src/planwise/component/maps.clj
@@ -16,6 +16,10 @@
   [{config :config}]
   (:mapserver-url config))
 
+(defn default-capacity
+  [{config :config}]
+  (:default-capacity config))
+
 (defn calculate-demand?
   [{config :config}]
   (boolean (:calculate-demand config)))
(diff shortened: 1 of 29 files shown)


src/planwise/component/maps.clj
```diff
@@ -21,7 +29,7 @@
   (digest/sha-256
     (str/join "_" (cons region-id polygons-with-capacities))))
 
-(defn- default-capacity
+(defn default-capacity
```
ggiraldez: `default-capacity` is defined twice with different bodies. Which one is the correct?

spalladino: Fixed in acada27842a3ccb0dbcee56d1341dc56af87874a

PR #223, opened 2016-09-14
Calculate demand in-memory

spalladino:
Demographics raster is loaded entirely in an array, and the resulting scaled byte-size raster is written directly to disk. Note that the program now requires an additional numeric param, saturation, used to scale the raster. This makes the saturate-demand script unnecessary.


cpp/calculate-demand.cpp
```diff
@@ -73,45 +59,79 @@ long readUnsatisfiedDemand(std::string targetFilename) {
...
-  GDALRasterBand* demoBand = demoDataset->GetRasterBand(1);
-  assert(demoBand->GetRasterDataType() == GDT_Float32);
-  int xBlockSize, yBlockSize;
-  demoBand->GetBlockSize(&xBlockSize, &yBlockSize);
-  GDALDataset* targetDataset = copyRaster(demoDataset, targetFilename, xBlockSize, yBlockSize);
-  GDALRasterBand* targetBand = targetDataset->GetRasterBand(1);
-  closeRaster(demoDataset);
+long calculateUnsatisfiedDemand(std::string targetFilename, std::string demoFilename, float saturation, std::vector<std::string> facilities, std::vector<float> capacities) {
+
+  #ifdef BENCHMARK
```
ggiraldez: Preprocessor directives are usually indented independently and not with the code https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives

cpp/calculate-demand.cpp
```diff
@@ -73,45 +59,79 @@ long readUnsatisfiedDemand(std::string targetFilename) {
...
+    demoBand->GetBlockSize(&xBlockSize, &yBlockSize);
+    demoDataset->GetGeoTransform(demoProjection);
+
+    targetXSize = demoDataset->GetRasterXSize();
+    targetYSize = demoDataset->GetRasterYSize();
+    targetNXBlocks = (targetXSize + xBlockSize - 1)/xBlockSize;
+    targetNYBlocks = (targetYSize + yBlockSize - 1)/yBlockSize;
+    nodata = demoBand->GetNoDataValue();
+    demoProjectionWKT = demoDataset->GetProjectionRef();
+    data = (float*) CPLMalloc(sizeof(float) * (xBlockSize * yBlockSize) * (targetNXBlocks * targetNYBlocks));
```
ggiraldez: I would `assert(data != NULL)` here just in case.

spalladino: I understand that [CPLMalloc](http://www.gdal.org/cpl__conv_8h.html#a9a806de98fbddb1337efdb18651aa0f7) raises if it fails.

cpp/calculate-demand.cpp
spalladino: Fixed in a24f045

PR #225, opened 2016-09-15
Try to fix isolated segments.

### LINKED ISSUE #12: Address isolated components in the topology graph
There are some instance in the OSM data where some ways are completely isolated from the road network (eg. a pedestrian pathway; see Lat -1.2595 Lon 36.8446). If a facility is located near one, the algorithm will choose a node in the isolated component for the origin of the isochrone and will yield an incorrect catchment area. There might be other places where there are disconnections due to human error when entering the map data as well.

Either remove the isolated components entirely or find a way to detect and correct these situations.

ggiraldez:
When importing the OSM data, as a post-processing step try to connect
the isolated segments back to the main network. The threshold used for
finding a nearby node is 5km which is the same as used when considering
a facility is too far during importing.

Closes #12

### DIFF
--- a/resources/planwise/plpgsql/fix-isolated.sql
+++ b/resources/planwise/plpgsql/fix-isolated.sql
@@ -28,8 +28,9 @@ END;
 $$ LANGUAGE plpgsql;
 
 
+DROP FUNCTION IF EXISTS connect_isolated_segments (float);
 CREATE OR REPLACE FUNCTION connect_isolated_segments (threshold float)
-RETURNS void AS $$
+RETURNS integer AS $$
 DECLARE
   segment RECORD;
   closest_to_source RECORD;
@@ -46,12 +47,22 @@ BEGIN
   created_ways := 0;
   fixed_segments := 0;
 
-  FOR segment IN SELECT w.*, a.the_geom AS source_geom, b.the_geom AS target_geom FROM ways w, ways_vertices_pgr a, ways_vertices_pgr b WHERE w.source = a.id AND a.cnt = 1 AND w.target = b.id AND b.cnt = 1 LOOP
+  CREATE OR REPLACE TEMP VIEW __isolated_ways AS
+         SELECT w.*, a.the_geom AS source_geom, b.the_geom AS target_geom
+         FROM ways w, ways_vertices_pgr a, ways_vertices_pgr b
+         WHERE w.source = a.id AND a.cnt = 1 AND w.target = b.id AND b.cnt = 1;
+
+  CREATE TEMP TABLE __isolated_nodes AS
+         SELECT source AS id FROM __isolated_ways
+         UNION ALL
+         SELECT target AS id FROM __isolated_ways;
+
+  FOR segment IN SELECT * FROM __isolated_ways LOOP
     isolated := isolated + 1;
     segment_fixed := FALSE;
     SELECT w.id, w.the_geom
            FROM ways_vertices_pgr w
-           WHERE id NOT IN (segment.source, segment.target)
+           WHERE id NOT IN (SELECT id FROM __isolated_nodes)
            ORDER by w.the_geom <-> segment.source_geom
            LIMIT 1
            INTO closest_to_source;
@@ -63,7 +74,7 @@ BEGIN
 
     SELECT w.id, w.the_geom
            FROM ways_vertices_pgr w
-           WHERE id NOT IN (segment.source, segment.target)
+           WHERE id NOT IN (SELECT id FROM __isolated_nodes)
            ORDER by w.the_geom <-> segment.target_geom
            LIMIT 1
            INTO closest_to_target;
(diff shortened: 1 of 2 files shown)


scripts/import-osm
```diff
@@ -54,6 +59,7 @@ ${OSM2PGROUTING} -f /tmp/$COUNTRY.osm -c $(dirname "${BASH_SOURCE}")/mapconfig.x
 echo "Populating ways_nodes table"
 psql -d $POSTGRES_DB -U $POSTGRES_USER -h $POSTGRES_HOST << SQL_SCRIPT
 SELECT populate_ways_nodes();
+SELECT connect_isolated_segments(5000);
```
spalladino: Is the function idempotent? If so, it might make sense to add it as a migration, so we don't forget running it manually on dev, stg or prod.

resources/planwise/plpgsql/fix-isolated.sql
```diff
@@ -0,0 +1,82 @@
...
+  fixed_segments INTEGER;
+  segment_fixed BOOLEAN;
+BEGIN
+  RAISE NOTICE 'Trying to fix up isolated segments';
+
+  isolated := 0;
+  created_ways := 0;
+  fixed_segments := 0;
+
+  FOR segment IN SELECT w.*, a.the_geom AS source_geom, b.the_geom AS target_geom FROM ways w, ways_vertices_pgr a, ways_vertices_pgr b WHERE w.source = a.id AND a.cnt = 1 AND w.target = b.id AND b.cnt = 1 LOOP
```
spalladino: This will only address isolated segments. Do we know if there are isolated components?

resources/planwise/plpgsql/fix-isolated.sql
```diff
@@ -0,0 +1,82 @@
...
+           WHERE id NOT IN (segment.source, segment.target)
+           ORDER by w.the_geom <-> segment.target_geom
+           LIMIT 1
+           INTO closest_to_target;
+    IF ST_Distance(segment.target_geom::geography, closest_to_target.the_geom::geography) < threshold THEN
+       new_way_gid := insert_way(segment.class_id, segment.maxspeed_forward, segment.target, closest_to_target.id);
+       created_ways := created_ways + 1;
+       segment_fixed := TRUE;
+    END IF;
+
```
spalladino: Instead of trying with `closest_to_source` first and then falling back to `closest_to_target`, I'd suggest querying both of them, and then taking the closest of the two (as long as it's within the threshold).

resources/planwise/plpgsql/fix-isolated.sql
```diff
@@ -0,0 +1,82 @@
...
+  geom := ST_MakeLine(ARRAY[source_node.the_geom, target_node.the_geom]);
+  length_m := ST_Length(geom::geography);
+
+  INSERT INTO ways
+         (class_id, length, length_m, source, target, x1, y1, x2, y2,
+          cost, one_way, maxspeed_forward, the_geom)
+         VALUES (class_id, ST_Length(geom), length_m, source, target,
+                 ST_X(source_node.the_geom), ST_Y(source_node.the_geom),
+                 ST_X(target_node.the_geom), ST_Y(target_node.the_geom),
+                 ST_Length(geom), 0, speed, geom)
```
spalladino: If we run this function independently, then it'd make sense to consider the traffic factor when inserting the cost here.

scripts/import-osm
ggiraldez: I don't like mixing this with migrations. If we add more countries we would run the function again to address isolated segments from the new data. It's also not truly idempotent: after run, we cannot change the threshold.

resources/planwise/plpgsql/fix-isolated.sql
ggiraldez: For that we'd need to pass the traffic factor to this function as well, which I don't like. Leaving the `cost_s` as NULL will trigger errors when importing new facilities that reach the new ways, which may be enough of a reminder. Would you negotiate printing an alert reminding you to re-run `apply_traffic_factor` after this function is executed?

spalladino: Worksforme!

resources/planwise/plpgsql/fix-isolated.sql
ggiraldez: There's not an easy way of finding out. And it can't be done automatically anyway. Think of roads on islands for example https://www.google.com.ar/maps/@-0.4662908,34.0303715,13.38z?hl=en

scripts/import-osm
spalladino: Ok, but we'll need to keep in mind to run this on all envs on the next release

resources/planwise/plpgsql/fix-isolated.sql
spalladino: I've just checked and `cnt` in my `ways_vertices_pgr table` is NULL for all rows. Isn't `pgr_analyzeGraph` needed to be run beforehand?

resources/planwise/plpgsql/fix-isolated.sql
ggiraldez: The code does attempt to connect both.

resources/planwise/plpgsql/fix-isolated.sql
ggiraldez: AFAIK, this is done by `osm2pgrouting`

ggiraldez: No, it's not. But we do it now in `import-osm` before running the connect function.

resources/planwise/plpgsql/fix-isolated.sql
ggiraldez: Done.

PR #233, opened 2016-09-19
Resume importer process

### LINKED ISSUE #93: Resume the import process if the process crashes
Either due to a redeploy or other reasons, the import process should be resilient and resume execution as soon as the server is up and running again.

spalladino:
Fixes #93

### DIFF
--- a/src/planwise/component/datasets.clj
+++ b/src/planwise/component/datasets.clj
@@ -45,6 +45,12 @@
   (->> (select-datasets-for-user (get-db store) {:user-id user-id})
        (map db->dataset)))
 
+(defn list-datasets-with-import-jobs
+  [store]
+  (->> (select-datasets-with-import-jobs (get-db store))
+       (map db->dataset)
+       (filter (comp some? :import-job))))
+
 (defn create-dataset!
   [store dataset]
   (let [db (get-db store)
--- a/src/planwise/component/datasets.clj
+++ b/src/planwise/component/datasets.clj
@@ -2,7 +2,8 @@
   (:require [com.stuartsierra.component :as component]
             [taoensso.timbre :as timbre]
             [hugsql.core :as hugsql]
-            [clojure.edn :as edn]))
+            [clojure.edn :as edn]
+            [planwise.util.hash :refer [update-if-contains]]))
 
 ;; ----------------------------------------------------------------------
 ;; Auxiliary and utility functions
@@ -18,13 +19,15 @@
   [record]
   (some-> record
           (update :import-result edn/read-string)
+          (update :import-job edn/read-string)
           (update :mappings edn/read-string)))
 
 (defn dataset->db
   [dataset]
   (some-> dataset
-          (update :import-result pr-str)
-          (update :mappings pr-str)))
+          (update-if-contains :import-result #(some-> % pr-str))
+          (update-if-contains :import-job #(some-> % pr-str))
+          (update-if-contains :mappings #(some-> % pr-str))))
 
 
 ;; ----------------------------------------------------------------------
--- a/src/planwise/component/datasets.clj
+++ b/src/planwise/component/datasets.clj
@@ -3,7 +3,7 @@
             [taoensso.timbre :as timbre]
             [hugsql.core :as hugsql]
             [clojure.edn :as edn]
-            [planwise.util.hash :refer [update-if-contains]]))
+            [planwise.util.hash :refer [update-if]]))
 
 ;; ----------------------------------------------------------------------
 ;; Auxiliary and utility functions
@@ -25,9 +25,9 @@
 (defn dataset->db
   [dataset]
   (some-> dataset
-          (update-if-contains :import-result #(some-> % pr-str))
-          (update-if-contains :import-job #(some-> % pr-str))
-          (update-if-contains :mappings #(some-> % pr-str))))
+          (update-if :import-result pr-str)
+          (update-if :import-job pr-str)
+          (update-if :mappings pr-str)))
 
 
 ;; ----------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


src/planwise/model/import_job.clj
```diff
@@ -390,6 +390,10 @@
                          :type-field type-field)]
     (import-job job-value)))
 
+(defn restore-job
+  [job]
+  (import-job (:state job) (:value job)))
```
ggiraldez: How does this cope with pending tasks, ie. tasks that were pending at the moment the job was serialized? After restoration, they won't get a succeeded/failed report, because they effectively don't exist, so the FSM can be left in a stuck state.

src/planwise/component/datasets.clj
```diff
@@ -42,6 +45,12 @@
   (->> (select-datasets-for-user (get-db store) {:user-id user-id})
        (map db->dataset)))
 
+(defn list-datasets-with-import-jobs
+  [store]
+  (->> (select-datasets-with-import-jobs (get-db store))
+       (map db->dataset)
+       (filter (comp some? :import-job))))
```
ggiraldez: Is this filter really necessary since you're querying for `import_job IS NOT NULL` already?

src/planwise/component/datasets.clj
```diff
@@ -18,13 +19,15 @@
...
           (update :mappings edn/read-string)))
 
 (defn dataset->db
   [dataset]
   (some-> dataset
-          (update :import-result pr-str)
-          (update :mappings pr-str)))
+          (update-if-contains :import-result #(some-> % pr-str))
+          (update-if-contains :import-job #(some-> % pr-str))
+          (update-if-contains :mappings #(some-> % pr-str))))
```
ggiraldez: I understand why this is necessary (the double guard `update-if-contains` and `some->`), but I don't like it much. Any way around it?

src/planwise/component/importer.clj
```diff
@@ -190,19 +191,38 @@
...
 
 ;; ----------------------------------------------------------------------------
+;; Jobs persistance
+
+(defn- persist-job
+  [component job]
+  (let [dataset-id (import-job/job-dataset-id job)]
+    (datasets/update-dataset (:datasets component)
+                             {:id dataset-id
+                              :import-job (dissoc job :fsm)})))
```
ggiraldez: I'd encapsulate the job internal structure in the `import-job` namespace and provide a `serialize-job` or similar. That way it's also symmetric with `restore-job` below.

src/planwise/component/datasets.clj
spalladino: How about changing it to a `(update-if :import-result pr-str)`, and `update-if` runs the update only if not-nil, instead of contains?

src/planwise/component/datasets.clj
spalladino: It's a leftover from a bug that caused finished jobs to be stored as `nil`, so it should not be necessary. Still, it is a safeguard; as such, I'd leave it there, though it should not be needed.

src/planwise/component/datasets.clj
ggiraldez: I like it.

src/planwise/model/import_job.clj
spalladino: The job is only serialised once a task is completed/failed, so if there are any pending tasks, these should be re-created when the job is restored and next-task is called. I can add a test for that to double-check.

ggiraldez: I mean tasks that are already executing. There can be up to `concurrent-workers` simultaneous (parallel) tasks.

spalladino: Yep, it fails when having multiple facilities processing tasks running at the same time, since `dispatch-process-facilities` is the only dispatch action that actually modifies the job. I'll have to review this then.

spalladino: Check 0a9d379f11420387a16be680de4b25e40fae9b5c for a possible solution

src/planwise/component/datasets.clj
spalladino: Fixed in 0509007e052aa704edbbc9a31bbef0659e7a0be0

src/planwise/component/importer.clj
spalladino: Fixed in d16b388e8871bac8db91ce4cfc1e20d3649b2e4e

src/planwise/model/import_job.clj
```diff
@@ -1,7 +1,8 @@
 (ns planwise.model.import-job
   (:require [reduce-fsm :as fsm]
             [schema.core :as s]
-            [clojure.core.match :refer [match]]))
+            [clojure.core.match :refer [match]]
+            [clojure.set :refer [rename-keys]]))
```
ggiraldez: This is no longer required.

test/planwise/component/importer_test.clj
```diff
@@ -75,15 +76,59 @@
...
+                                                   :page 1, :user-ident user-ident,
+                                                   :facility-ids [1 2 3 4 5 6],
+                                                   :facility-count 6})})]
+        (swap! (:jobs importer) assoc 1 job)
+
+        ; Fire 4 concurrent process facility tasks
+        (let [tasks (repeatedly 4 #(taskmaster/next-task importer))]
+          (dotimes [n 4]
+            (is (= [1 [:process-facilities [(inc n)]]]
+                   (:task-id (nth tasks n)))))
```
ggiraldez: You can also generate the desired result and check for equality. Ie.
```clojure
(let [tasks (repeatedly 4 #(taskmaster/next-task importer))
      expected-task-ids (for [i (range 4)] [:process-facilities [(inc i)]])]
  (is (= (map :task-id tasks) expected-task-ids)))
```

src/planwise/model/import_job.clj
spalladino: Fixed in 3e779a3f4a71735ac26a878641ff72183b7f63ce

test/planwise/component/importer_test.clj
spalladino: True, using `dotimes` feels less _functional-ly_. Fixed in b2d0a82777c91af43543487d023d8a8284cdc533.

PR #234, opened 2016-09-20
Remember wizard state for projects

### LINKED ISSUE #190: Remember wizard state when the user navigates away from a project
When the user creates a new project, the wizard mode is activated for it until she visits all the project tabs. If she navigates away from the project and there are still tabs to visit, the next time she opens the project she should be taken back to the spot where she left off.

ggiraldez:
Fixes #190 

- Add a state field to the projects table
- Have the client send the wizard state after each navigation and save
  it to the field
- When loading a project, retrieve the state field and restore the
  wizard to the state it was before navigating away from the project


src/planwise/client/current_project/handlers.cljs
```diff
@@ -54,40 +54,48 @@
...
+
+(defn- update-wizard-state
+  [db next-tab]
+  (let [state (:wizard db)
+        active? (:set state)]
+    (if active?
+      (let [project-id (db/project-id db)
+            new-state (visit-tab state next-tab)]
+        (when (not= state new-state)
+          (api/update-project-state project-id new-state (fn [data] :ignore)))
```
spalladino: Are handlers mandatory? Why not just remove the last parameter?

src/planwise/component/projects.clj
```diff
@@ -132,9 +134,11 @@
   (jdbc/with-db-transaction [tx (get-db service)]
     (let [project-id (:id project)
           goal       (:goal project)
+          state      (:state project)
           filters    (:filters project)
           params     {:project-id project-id
                       :goal       goal
+                      :state      (some-> state pr-str)
                       :filters    (some-> filters pr-str)}
           result     (update-project* tx params)]
```
spalladino: Why not use `project->db` here? `update-if` might come in handy :P

PR #235, opened 2016-09-21
Handle polygons without population in calculate demand

spalladino:
If there was an error during the import process, the population
for a polygon or a polygon-region might be null. If so, simply
skip those in the calculate demand computation instead of failing
with a null reference exception.


src/planwise/component/maps.clj
```diff
@@ -73,6 +73,8 @@
   [service region-id facilities]
   (->> facilities
     (filter :polygon-id)
+    (filter :population)
```
ggiraldez: A value of `0` will pass the filter (since it's truthy) but will lead to a division by zero error later.

PR #251, opened 2016-09-23
Process unprocessed facilities on a dataset reimport

### LINKED ISSUE #250: Updating a dataset for which the import was cancelled doesn't reprocess
When updating a dataset which has facilities for which the isochrones were not computed, the system should reprocess such facilities, regardless of whether the data from Resourcemap changed or not.

spalladino:
If a dataset is cancelled and then reimported, ensure those facilities that were not processed in the first run are indeed processed.

Change the return type of insert-facilities so it returns a list of all the new facilities inserted, along with an inserted-status attribute that yields whether it's new, moved, existing or updated.

Fixes #250

### DIFF
--- a/src/planwise/component/facilities.clj
+++ b/src/planwise/component/facilities.clj
@@ -44,8 +44,8 @@
 
 (defn insert-facilities!
   "Upserts the facilities, identifying them by dataset-id and site-id,
-   returning a hash of state into list of ids, where state is one of
-   :existing, :updated, :moved or :new."
+   returning a list of the facilities with their ids, plus a new
+   :insertion-status, where state is one of :existing, :updated, :moved or :new."
   [service dataset-id facilities]
   (jdbc/with-db-transaction [tx (get-db service)]
     (let [updateable-keys [:name :type-id :capacity]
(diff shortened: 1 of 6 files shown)


src/planwise/component/facilities.clj
```diff
@@ -44,8 +44,8 @@
 
 (defn insert-facilities!
   "Upserts the facilities, identifying them by dataset-id and site-id,
-   returning a hash of state into list of ids, where state is one of
-   :existing, :updated, :moved or :new."
+   returning a list of the facilities with there ids, plus a new
```
ggiraldez: there -> their

src/planwise/component/facilities.clj
```diff
@@ -66,30 +66,32 @@
                     ; Insert new record if no existing facility with the site id was found
                     (nil? existing)
                     (let [result (insert-facility! tx (assoc facility :dataset-id dataset-id))]
-                      [:new (:id result)])
+                      (-> facility
+                        (assoc :dataset-id dataset-id
+                               :id (:id result)
+                               :insertion-status :new)
+                        (update :processing-status identity)))
```
ggiraldez: Why do you need this `update`?

PR #303, opened 2018-03-21
Fix project for valid spec

solsarratea:


src/planwise/endpoint/projects2.clj
```diff
@@ -14,14 +14,27 @@
 (timbre/refer-timbre)
 
 
+(defn- default
+[map path value-default]
+(let [value (get-in map path)]
+    (if (nil? value) 
+      (assoc-in map path value-default)
+      (assoc-in map path value))))
```
bcardiff: This is the same as `(assoc-in map path value)` is equal to `map` here

src/planwise/endpoint/projects2.clj
```diff
@@ -14,14 +14,27 @@
 (timbre/refer-timbre)
 
 
+(defn- default
+[map path value-default]
```
bcardiff: missing indentation

src/planwise/endpoint/projects2.clj
```diff
@@ -34,10 +47,11 @@
       (response project)))
 
    (PUT "/:id" [id project :as request]
-    (println project)
-     (assert (s/valid? ::project project) "invalid project-data")
-     (let [user-id    (util/request-user-id request)]
-       (projects2/update-project service id project)))
+     (let [user-id    (util/request-user-id request)
+           id         (:id project)
+           valid-project    (assoc project :config (apply-default (:config project)))]
```
bcardiff: Use the `apply-default` in the `GET "/:id"`, but not in the `PUT "/:id"` route. So the client receives a full config.

PR #313, opened 2018-03-28
Start project

### LINKED ISSUE #296: Add a Start action for a project
In the config of a project, allow the user to "start" the project.
This means a confirmation of all configuration and the beginning of the search of scenarios.
Add a `state` property to the projects.

When created, a project is in `draft` mode. Where the user is able to edit all the configuration.
Once started the project will begin to compute the scenarios (currently nothing happen)
Eventually the project will become to `done` state.

Also when started, store the current dataset version in the project.

ftarulla:
Fix #296


src/planwise/client/projects2/api.cljs
```diff
@@ -27,3 +27,8 @@
   {:method    :get
    :params    {:id project-id}
    :uri       (str "/api/projects2/" project-id)})
+
+(defn- start-project!
+  [project-id]
+  {:method    :post
+   :uri       (str "/api/projects2/start/" project-id)})
```
bcardiff: The route should be `/api/projects2/:id/start`

PR #331, opened 2018-04-04
Use cmake for building binaries

ggiraldez:
- Use cmake to build the C++ binaries
- Compile in different directories depending on the OS/arch
- Provide trampolines in place of the binaries to select the appropriate binary depending on the running platform (for backwards compatibility with existing scripts)
- Silence compilation warnings


bcardiff:
Rebased on master to exclude cherrypicked commit.

@ggiraldez on is supposed to set `BIN_PATH` manually when developing local?

ggiraldez:
No, you shouldn't need to. The default is `cpp/` as set in https://github.com/instedd/planwise/blob/ci/cmake/dev/resources/dev.edn#L36 and the trampolines are there.

PR #337, opened 2018-04-11
Compute initial scenario

### LINKED ISSUE #297: Compute the initial scenario when the project is started

ggiraldez:
Closes #297 

- Implement function primitives to read and write raster files
- Implement function primitives to perform computations over rasters (in Java for performance)
- Integrate Virgil to dynamically recompile and reload Java classes during development
- Compute the initial scenario for a project using the base rasters, project configuration and affected sites coverages
- Extract the interface of JobRunner into a boundary
- Spawn the computation asynchronously using the JobRunner when starting a project

### DIFF
--- a/resources/planwise/sql/scenarios.sql
+++ b/resources/planwise/sql/scenarios.sql
@@ -46,6 +46,14 @@ UPDATE scenarios
 WHERE
   id = :id
 
+-- :name db-update-scenario-state! :! :1
+UPDATE "scenarios"
+  SET "raster" = :raster,
+      "demand-coverage" = :demand-coverage,
+      "state" = :state,
+      "updated-at" = NOW()
+  WHERE "id" = :id;
+
 -- :name db-list-scenarios-names :?
 SELECT name
 FROM scenarios
--- a/src/planwise/component/scenarios.clj
+++ b/src/planwise/component/scenarios.clj
@@ -42,15 +42,16 @@
       (update :changeset edn/read-string)))
 
 (defn create-initial-scenario
-  [store project-id]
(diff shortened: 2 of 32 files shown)


resources/planwise/sql/scenarios.sql
```diff
@@ -46,6 +46,14 @@ UPDATE scenarios
 WHERE
   id = :id
 
+-- :name db-update-scenario-state! :! :1
+UPDATE "scenarios"
+  SET "raster" = :raster,
+      "demand-coverage" = :demand-coverage,
+      "state" = :state,
+      "updated-at" = NOW()
```
bcardiff: Changing updated-at should be done `db-update-scenario!` and `db-create-scenario!`.

I actually wouldn't mind changing it in `db-update-scenario-state!` since I would expect to be an "updated by user". So we later can use it to dismiss running background jobs.

src/planwise/component/scenarios.clj
```diff
@@ -40,15 +42,33 @@
...
-  (db-create-scenario! (get-db store)
-                       {:name "Initial"
-                        :project-id project-id
-                        :investment 0
-                        :demand-coverage nil
-                        :changeset "[]"
-                        :label "initial"}))
+  [store project]
+  (let [project-id  (:id project)
+        scenario-id (:id (db-create-scenario! (get-db store)
```
bcardiff: db-create-scenario should set the initial state to some value for scheduled or running

bcardiff:
Other than those two minor comments LGTM and we can move forward. Fix and merge or merge and defer 👍

PR #343, opened 2018-04-13
Compute color scale using demand quartiles and render appropriate map

### LINKED ISSUE #338: Create color scale population / demand layers
From a tiff with floats build
* compute min,q-25%,med,q-75%,max values
* build a raster layer with 5 color scale de mapserver use

ggiraldez:
Closes #338

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,7 +55,7 @@ services:
     # dockerfile: Dockerfile.mapserver
     volumes:
       - "./data:/data:ro"
-      - "./mapserver/planwise.map:/etc/mapserver/mapserver.map:ro"
+      - "./mapserver/planwise.map:/etc/mapserver/planwise.map:ro"
     ports:
       - "5001:80"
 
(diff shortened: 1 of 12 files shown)


bcardiff:
@ggiraldez some followup regarding the cache? Does the new image support changing the file or if needed to be updated a new file name should be used to avoid hitting the cache?

Also, should those images be used in the server and the Dockerfile in mapserver folder wont be needed in the future?

bcardiff:
@ggiraldez another doubt is if whether or not the datafile passed to wms-tile-layer should have the .map appended of the .map.tif files or it that is added automatically.

ggiraldez:
@bcardiff 

- A new filename should be used to bypass caching from MapCache.
- Docker images for staging/production should be rebuilt with the updated base versions and configuration files. I'm not really sure where are the original Dockerfile(s) used to build them in the first place, as the images contain more files than those specified en `mapserver/Dockerfile.*` But I think those should be fine after updating the base version.
- Mapserver will only add `.tif` so yes, the `.map` should be passed in the parameter.

PR #356, opened 2018-04-18
Feature/354 initial scenario read only

### LINKED ISSUE #353: Define wording for sidebar of initial scenario
the current one does not make sense since there is no increase with respect the initial scenario ;-)

check with @jkicillof

### LINKED ISSUE #354: Make the initial scenario read only from the API and UI
Currently the initial scenario can be renamed

solsarratea:
Closes #354 
Closes #353

### DIFF
--- a/src/planwise/component/scenarios.clj
+++ b/src/planwise/component/scenarios.clj
@@ -94,11 +94,12 @@
 
 (defn update-scenario
   [store project {:keys [id name changeset]}]
-  ;; TODO fail if updating initial. initial scenario should be readonly
   ;; TODO assert scenario belongs to project
-  (assert (s/valid? ::model/change-set changeset))
   (let [db (get-db store)
-        project-id (:id project)]
+        project-id (:id project)
+        label (:label (get-scenario store id))]
+    (assert (s/valid? ::model/change-set changeset))
+    (assert (not= label "initial"))
     (db-update-scenario! db
                          {:name name
                           :id id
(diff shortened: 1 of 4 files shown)


src/planwise/component/scenarios.clj
```diff
@@ -93,25 +93,26 @@
...
-      (db-update-scenarios-label! db {:project-id project-id})
-      (jr/queue-job (:jobrunner store)
-                    [::boundary/compute-scenario id]
-                    {:store store
-                     :project project}))))
+  (assert (s/valid? ::model/change-set changeset)
+          (let [db (get-db store)
+                project-id (:id project)
+                label (:label (get-scenario store id))]
+            (when (not= label "initial")
```
bcardiff: assert can be used here to ensure the label is not initial.

test/planwise/component/scenarios_test.clj
```diff
@@ -121,11 +121,13 @@
...
+                              :label "sub-optimal"
+                              :investment 10000)]
       (do
-        (scenarios/update-scenario store project scenario)
-        (let [updated-scenario (scenarios/get-scenario store scenario-id)]
-          (is (some? (:updated-at updated-scenario)))
-          (is (:changeset updated-scenario) [])
-          (is (:investment updated-scenario) (:investment scenario))
-          (is (:state scenario) (:state updated-scenario)))))))
+        (scenarios/update-scenario store project new-scenario)
```
bcardiff: if assert is used to validate label is not initial, then here
`(is (thrown? java.lang.AssertionError (scenarios/update-scenario store project new-scenario)))`
is probably needed.

ref: http://clojure.github.io/clojure/clojure.test-api.html and https://clojuredocs.org/clojure.core/assert

PR #374, opened 2018-04-24
Create handler to prevent block scrolling

### LINKED ISSUE #360: Check unable to scroll after a project is deleted
The class is still applied mdc-dialog-scroll-lock
Maybe the action of the dialog should be deferred after the accept is performed

solsarratea:
Fixes #360


src/planwise/client/dialog.cljs
```diff
@@ -1,11 +1,12 @@
 (ns planwise.client.dialog
   (:require [planwise.client.ui.rmwc :as m]
+            [re-frame.core :refer [dispatch]]
             [planwise.client.utils :as utils]))
 
 (defn new-dialog
```
ggiraldez: We might want to rename this component if it's used for other purposes other than creating a new thing. In particular if it's used for deleting such things ;)

src/planwise/client/dialog.cljs
```diff
@@ -1,11 +1,12 @@
...
-  [{:keys [open? title content accept-fn cancel-fn delete-fn]}]
+  [{:keys [open? title content delete-dispatch accept-fn cancel-fn delete-fn]}]
   [m/Dialog {:open open?
-             :on-accept accept-fn
+             :on-accept (or  accept-fn #(dispatch [:dialog/delay-dispatch delete-dispatch]))
```
ggiraldez: I don't like mix and matching callback style attributes with dispatch vectors. For generality, I'd prefer to just use `accept-fn` and make the dispatch function at the use point.

src/planwise/client/handlers.cljs
```diff
@@ -76,3 +76,8 @@
  (fn [_ [_ time]]
    (when (= 0 (mod time 1000))
      {:dispatch [:datasets/refresh-datasets time]})))
+
+(rf/reg-event-fx
+ :dialog/delay-dispatch
+ (fn [_ [_ event-vector]]
+   {:dispatch-later [{:ms 150 :dispatch event-vector}]}))
```
ggiraldez: See my comment in #360. There may be a better workaround for this particular bug. Otherwise, please add a comment here explaining why the delayed dispatch.

src/planwise/client/projects2/components/common.cljs
```diff
@@ -0,0 +1,11 @@
+(ns planwise.client.projects2.components.common
+  (:require [planwise.client.dialog :refer [new-dialog]]
+            [planwise.client.components.common2 :as common2]))
+
+(defn delete-project-dialog
+  [state id]
+  [new-dialog {:open? @state
```
ggiraldez: It's more idiomatic to just pass the value of the `state` atom. You'd need to extract the cancel callback at the use point as well, but that change will decouple the component from the use case, which is a good thing.

src/planwise/client/projects2/components/settings.cljs
```diff
@@ -93,9 +93,9 @@
            [current-project-input "Budget" [:config :actions :budget] valid-input]]
           [:div
            [project-start-button {} @current-project]
-           [project-delete-button delete?]]
+           [project-delete-button delete? (:id @current-project)]]
           [new-dialog {:open? @delete?
                        :title "Delete Project"
-                       :accept-fn #(dispatch [:projects2/delete-project (:id @current-project)])
+                       :delete-dispatch [:projects2/delete-project (:id @current-project)]
```
ggiraldez: See comment above regarding the mix of `-fn` callbacks and dispatch vectors.

src/planwise/client/projects2/components/common.cljs
```diff
@@ -0,0 +1,12 @@
+(ns planwise.client.projects2.components.common
+  (:require [planwise.client.dialog :refer [dialog]]
+            [re-frame.core :refer [dispatch]]
+            [planwise.client.components.common2 :as common2]))
+
+(defn delete-project-dialog
+  [{:keys [open? id cancel-fn delete-fn]}]
+  [dialog {:open? open?
+           :title "Delete Project"
+           :accept-fn delete-fn
```
solsarratea: It was redundant to have an "OK" and "DELETE" button in the delete dialog

src/planwise/client/projects2/components/settings.cljs
```diff
@@ -67,53 +68,51 @@
...
+             [:section {:class-name "project-settings-section"}
+              [section-header 5 "Actions"]
+              [current-project-input "Budget" [:config :actions :budget] valid-input]]]]]
+          [:div {:class-name "project-settings-actions"}
+           [project-delete-button delete?]
+           [project-start-button {} @current-project]]]
+         [delete-project-dialog {:open? @delete?
+                                 :cancel-fn hide-dialog
+                                 :delete-fn (fn []
+                                             ;(hide-dialog)
```
ggiraldez: Is this a leftover? Are you sure we don't need to explicitly hide the dialog here?

src/planwise/client/scenarios/edit.cljs
```diff
@@ -52,10 +52,10 @@
...
-                   :content (changeset-dialog-content @site)
-                   :delete-fn #(dispatch [:scenarios/delete-site @site-index])
-                   :accept-fn  #(dispatch [:scenarios/accept-changeset-dialog])
-                   :cancel-fn  #(dispatch [:scenarios/cancel-changeset-dialog])}))))
+      (dialog {:open? (= @view-state :changeset-dialog)
+               :title "Edit Site"
+               :content (changeset-dialog-content @site)
+               :delete-fn #(dispatch [:scenarios/delete-site @site-index])
+               :accept-fn  #(dispatch [:scenarios/accept-changeset-dialog])
+               :cancel-fn  #(dispatch [:scenarios/cancel-changeset-dialog])}))))
```
ggiraldez: Nit-pick: fix the spacing please :)

src/planwise/client/projects2/components/settings.cljs
solsarratea: I should have commented that when the accept dialog button is not being used (in this current version we just use delete), it is also not necessary to use hide-dialog

PR #384, opened 2018-05-07
Feature/319 filter sites via tags

solsarratea:


src/planwise/client/projects2/handlers.cljs
```diff
@@ -152,3 +152,27 @@
...
+
+(rf/reg-event-fx
+ :projects2/save-tag
+ in-projects2
+ (fn [{:keys [db]} [_ tag]]
+   (let [path [:current-project :config :sites :tags]
+         n (count (get-in db path))]
+     {:db          (if (pos? n)
+                     (assoc-in db (into path [n]) tag)
+                     (assoc-in db path [tag]))
```
ggiraldez: The updated db value can be simplified to `(update-in db [:current-project :config :sites :tags] (comp vec conj) tag)`

src/planwise/client/projects2/handlers.cljs
```diff
@@ -152,3 +152,27 @@
...
+      :dispatch [:projects2/persist-current-project]})))
+
+(rf/reg-event-fx
+ :projects2/delete-tag
+ in-projects2
+ (fn [{:keys [db]} [_ index]]
+   (let [path [:current-project :config :sites :tags]
+         tags* (vec (keep-indexed #(if (not= %1 index) %2) (get-in db path)))]
```
ggiraldez: This (remove vector element) looks useful enough to extract as a function into `planwise.client.utils`.

src/planwise/component/datasets2.clj
```diff
@@ -57,6 +58,11 @@
   (db-find-sites (get-db store) {:dataset-id dataset-id
                                  :version version}))
 
+(defn filter-sites-by-tags
+  [sites tags]
+  (reduce
+   (fn [sites tag] (let [filtered-sites  (filter #(-> % :tags (includes? tag)) sites)] filtered-sites)) sites tags))
```
ggiraldez: This will not work correctly for tags which are a substring of another tag. Ie. `includes?` will return `true` for "foo" in the string "foobar baz".

test/planwise/component/datasets2_test.clj
```diff
@@ -85,3 +113,19 @@
           listed-sites             (datasets2/sites-by-version store dataset-id last-version-dataset)]
       (is (= (count listed-sites) 2))
       (is (= last-version-dataset 2)))))
+
+;; ----------------------------------------------------------------------
+;; Testing site's tag filtering
+
+(deftest filtering-sites
+  (test-system/with-system (test-config fixture-filtering-sites-tags)
```
ggiraldez: This is ok, but you don't need to spin up a test system to test the filtering functionality. Since it's just a side-effect free function with no real dependencies, you can construct the sites collection and use it to invoke `planwise.component.datasets2/filter-sites-by-tags` directly

PR #387, opened 2018-05-07
Show project settings (read-only) when started

### LINKED ISSUE #363: Show project information when started
Everything should be read only, except the name that should be editable and with autosave.

The name in the list of project should be in sync without further coding, but double check the behavior.

ftarulla:
Fix #363

### DIFF
--- a/src/planwise/client/projects2/components/dashboard.cljs
+++ b/src/planwise/client/projects2/components/dashboard.cljs
@@ -57,8 +57,8 @@
       [:th {:class "col3"} (str (get-in current-project [:config :demographics :unit-name]) " coverage")]
       [:th {:class "col4"} "Investment"]
       [:th {:class "col5"} "Actions"]]]
-
-    (into [:tbody] (map #(scenarios-list-item (:id current-project) %) scenarios))]])
+    [:tbody
+     (map #(scenarios-list-item (:id current-project) %) scenarios)]]])
 
 (defn- project-settings
   []
--- a/src/planwise/client/projects2/components/dashboard.cljs
+++ b/src/planwise/client/projects2/components/dashboard.cljs
@@ -18,8 +18,8 @@
              :on-change (fn [evt]
                           (let [tab-index (.-value (.-target evt))]
                             (cond
-                              (= tab-index 0) (println "Load scenarios!!")
-                              (= tab-index 1) (println "Load settings!!"))))}
+                              (= tab-index 0) (dispatch [:projects2/project-scenarios])
+                              (= tab-index 1) (dispatch [:projects2/project-settings]))))}
    [m/Tab "Scenarios"]
    [m/Tab "Settings"]])
 
@@ -70,7 +70,7 @@
         hide-dialog (fn [] (reset! delete? false))
         id (:id @current-project)
         scenarios-sub (rf/subscribe [:scenarios/list])]
-    (fn []
+    (fn [active-tab]
       (let [scenarios (asdf/value @scenarios-sub)]
         (when (asdf/should-reload? @scenarios-sub)
           (rf/dispatch [:scenarios/load-scenarios]))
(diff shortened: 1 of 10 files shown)


src/planwise/client/projects2/components/dashboard.cljs
```diff
@@ -8,11 +8,19 @@
...
             [planwise.client.ui.rmwc :as m]
-            [planwise.client.utils :as utils]))
+            [planwise.client.utils :as utils]
+            [planwise.client.projects2.components.settings :as settings]))
 
 (defn- project-tabs
-  [{:keys [active] :or {active 0}}]
-  [m/TabBar {:activeTabIndex active}
+  [{:keys [active] :or {active :scenarios}}]
+  [m/TabBar {:activeTabIndex (cond
```
ggiraldez: You may want to use `case` here instead (when all the conditions are simple equals), or even a map, which can act as functions on the keys:

```clojure
({:scenarios 0 :settings 1} active)
```

src/planwise/client/projects2/components/dashboard.cljs
```diff
@@ -8,11 +8,19 @@
...
+                               (= active :scenarios) 0
+                               (= active :settings) 1)
+             :on-change (fn [evt]
+                          (let [tab-index (.-value (.-target evt))]
+                            (cond
```
ggiraldez: Ditto for `case` vs. `cond`. Or `dispatch` on the mapping of `tab-index` into the handler names (careful with not dispatching `nil` though).

src/planwise/client/projects2/components/dashboard.cljs
```diff
@@ -68,10 +80,14 @@
...
                                  {:title (:name @current-project)
-                                  :tabs [project-tabs {:active 0}]
+                                  :tabs [project-tabs {:active active-tab}]
                                   :secondary-actions (project-secondary-actions @current-project delete?)})
            [delete-project-dialog {:open? @delete?
                                    :cancel-fn hide-dialog
-                                   :delete-fn #(rf/dispatch [:projects2/delete-project id])}]])))))
+                                   :delete-fn #(rf/dispatch [:projects2/delete-project id])}]
+           [ui/panel {}
+            (cond
```
ggiraldez: Ditto on `case` vs. `cond`.

src/planwise/client/projects2/components/settings.cljs
```diff
@@ -60,16 +63,17 @@
              :on-click #(reset! state true)} "Delete"])
 
 (defn- tag-chip
-  [props index input]
+  [props index input read-only]
   [m/Chip props [m/ChipText input]
-   [m/ChipIcon {:use "close"
-                :on-click #(dispatch [:projects2/delete-tag index])}]])
+   (cond
```
ggiraldez: Use `when-not` for this case.

src/planwise/client/projects2/components/settings.cljs
```diff
@@ -87,54 +91,72 @@
...
+          [datasets-dropdown-component {:label     "Dataset"
+                                        :value     (:dataset-id @current-project)
+                                        :on-change #(dispatch [:projects2/save-key :dataset-id %])
+                                        :disabled? read-only}]
+
+          [current-project-input "Capacity workload" [:config :sites :capacity] valid-input {:disabled read-only}]
+          [m/TextFieldHelperText {:persistent true} (str "How many " (get-in @current-project [:config :demographics :unit-name]) " can be handled per site capacity")]
+
+          (cond
+            (false? read-only) [tag-input])
```
ggiraldez: Use `when-not` here as well.

src/planwise/client/projects2/views.cljs
```diff
@@ -37,5 +37,7 @@
         (let [section      (:section @page-params)]
           (cond
```
ggiraldez: Please replace with a `case` while you are at it :)

PR #390, opened 2018-05-10
Filter projects by owner

### LINKED ISSUE #386: I can access Projects that do not belong to me.
Found in version: 2428ec6

Preconditions:

- There is a Project `P1`.
- There are two Users (`U1` and `U2`).
- `P1` belongs to the User `U1`.

Steps to reproduce:

1. Sign in as `U2`.
2. Access `P1` through URL.

Obtained result:

- `U2` can access `P1`.

Expected:

- Projects should be filtered by User.

ftarulla:
Fix #386

### DIFF
--- a/src/planwise/endpoint/projects2.clj
+++ b/src/planwise/endpoint/projects2.clj
@@ -18,6 +18,14 @@
   [project]
   (dissoc project :deleted-at))
 
+(defn- filter-projects-owned-by
+  [projects owner-id]
+  (filter (fn [proj] (= (:owner-id proj) owner-id)) projects))
+
+(defn- filter-owned-by
+  [project owner-id]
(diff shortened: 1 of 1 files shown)


src/planwise/endpoint/projects2.clj
```diff
@@ -18,6 +18,14 @@
   [project]
   (dissoc project :deleted-at))
 
+(defn- filter-projects-owned-by
+  [projects owner-id]
+  (filter (fn [project] (= (:owner-id project) owner-id)) projects))
+
+(defn- filter-owned-by
+  [project owner-id]
+  (first (filter-projects-owned-by [project] owner-id)))
```
ggiraldez: This is backwards :) You should use build a predicate for a single project and use that in the case of collections.

PR #392, opened 2018-05-10
Feature/319 continuation filter by tags

solsarratea:
Closes #384

### DIFF
--- a/resources/planwise/sql/datasets2.sql
+++ b/resources/planwise/sql/datasets2.sql
@@ -48,6 +48,12 @@ SELECT s2.id, s2.name, s2.lat, s2.lon, s2.capacity, s2.type, s2.tags, s2c.raster
           AND s2.tags::tsvector @@ :tags::tsquery;
          /*~ ) ~*/;
 
+-- :name db-count-sites-with-tag :? :1
+SELECT COUNT(*) FROM sites2
+    WHERE "dataset-id" = :dataset-id
+    AND version = :version
+    AND tags::tsvector @@ :tag::tsquery;
+
 -- :name db-enum-site-ids :?
 SELECT "id" FROM "sites2"
        WHERE "dataset-id" = :dataset-id
--- a/resources/planwise/sql/datasets2.sql
+++ b/resources/planwise/sql/datasets2.sql
@@ -48,12 +48,13 @@ SELECT s2.id, s2.name, s2.lat, s2.lon, s2.capacity, s2.type, s2.tags, s2c.raster
           AND s2.tags::tsvector @@ :tags::tsquery;
          /*~ ) ~*/;
 
--- :name db-count-sites-with-tag :? :1
-SELECT COUNT(*) FROM sites2
+-- :name db-count-sites-with-tags :? :1
+SELECT COUNT(*) FROM sites2 s2
     WHERE "dataset-id" = :dataset-id
     AND version = :version
-    /*~ (if (seq (:tag params)) */
-    AND tags::tsvector @@ :tag::tsquery;
+    AND s2."the_geom" @ (SELECT "the_geom" FROM regions WHERE id = :region-id)
+    /*~ (if (seq (:tags params)) */
+    AND s2.tags::tsvector @@ :tags::tsquery;
     /*~ ) ~*/;
 
 -- :name db-enum-site-ids :?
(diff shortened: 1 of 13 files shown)


resources/planwise/sql/datasets2.sql
```diff
@@ -43,7 +43,18 @@ SELECT s2.id, s2.name, s2.lat, s2.lon, s2.capacity, s2.type, s2.tags, s2c.raster
...
-          AND s2c.options = :options;
+          AND s2c.options = :options
+         /*~ (if (seq (:tags params)) */
+          AND s2.tags::tsvector @@ :tags::tsquery;
+         /*~ ) ~*/;
+
+-- :name db-count-sites-with-tag :? :1
+SELECT COUNT(*) FROM sites2
+    WHERE "dataset-id" = :dataset-id
+    AND version = :version
```
ggiraldez: For consistent results, please add the region clipping filter, ie.
```
AND s2."the_geom" @ (SELECT "the_geom" FROM regions WHERE id = :region-id)
```
Of course, this needs to be conditional to the `:region-id` parameter not being `nil`.

src/planwise/client/projects2/components/settings.cljs
```diff
@@ -85,6 +85,10 @@
                     :on-change #(reset! value (-> % .-target .-value))
                     :value @value}])))
 
+(defn- count-sites
+  [tags {{:keys [total filtered]} :dataset-sites}]
+  [:p "Available sites: " (if (empty? tags) total filtered) " / " total])
```
ggiraldez: The label here would be "Usable sites" or "Selected sites". It should also display regardless of whether tags have been selected. Ie. with no tags, all sites within the region will be used.

src/planwise/component/engine.clj
```diff
@@ -43,10 +44,10 @@
         tags             (get-in config [:sites :tags])
         filter-options   {:region-id          region-id
                           :coverage-algorithm coverage-algorithm
-                          :coverage-options   coverage-options}
+                          :coverage-options   coverage-options
+                          :tags tags}
```
ggiraldez: You need to build the tsquery from the tags here for PostgreSQL.

src/planwise/component/datasets2.clj
```diff
@@ -197,13 +192,25 @@
   (let [db-spec   (get-db store)
         region-id (:region-id filter-options)
         algorithm (:coverage-algorithm filter-options)
-        options   (:coverage-options filter-options)]
+        options   (:coverage-options filter-options)
+        tags      (str/join "& " (:tags filter-options))]
```
ggiraldez: Are there any escaping issues with the user-provided tags? You may want to extract the tsquery builder function into some utils namespace and re-use it in the engine component below.

src/planwise/endpoint/projects2.clj
```diff
@@ -31,7 +31,8 @@
    (PUT "/:id" [id project :as request]
      (let [user-id  (util/request-user-id request)
            id       (Integer. id)
-           project  (assoc project :id id)] ;; honor id of route
+           project  (-> project (assoc :id id)
+                        (dissoc :dataset-sites))] ;; honor id of route
```
ggiraldez: This shouldn't be necessary. Eventually, the projects component should filter out invalid/unknown attributes. Extra attributes shouldn't matter for the spec either.

src/planwise/component/datasets2.clj
```diff
@@ -197,13 +192,25 @@
...
                                                     :algorithm  algorithm
-                                                    :options    (some-> options pr-str)})))
-
+                                                    :options    (some-> options pr-str)
+                                                    :tags tags})))
+
+(defn count-sites-filter-by-tag
+  [store dataset-id tag]
+  (let [db-spec  (get-db store)
+        count-fn (fn [db-spec tag]
```
ggiraldez: Since the `count-fn` is defined in the let binding, you don't need to pass `db-spec` as an argument. It is included in the closure scope.

src/planwise/component/datasets2.clj
```diff
@@ -197,13 +192,25 @@
...
+                   (let [{:keys [last-version]} (get-dataset store dataset-id)]
```
ggiraldez: `version` should probably be an optional parameter to the function. That way we can reuse this for counting sites for "older" projects.

PR #398, opened 2018-05-14
Validate project settings before start

### LINKED ISSUE #393: Validate that the project settings before starting
There's no check whether the project settings are valid before attempting to start the project. Check that all necessary attributes are filled in before enabling the Start operation.

ftarulla:
Fix #393

### DIFF
--- a/src/planwise/client/projects2/components/settings.cljs
+++ b/src/planwise/client/projects2/components/settings.cljs
@@ -48,11 +48,49 @@
                    :value     value
                    :disabled  disabled}])))
 
+(defn- ^boolean valid-goal?
+  [project]
+  (every? false? [(blank? (:name project))
+                  (nil? (:region-id project))]))
+
+(defn- ^boolean valid-demographics?
+  [project]
+  (let [demographics (get-in project [:config :demographics])]
+    (every? false? [(nil?   (:population-source-id project))
+                    (blank? (:unit-name demographics))
+                    (blank? (:target demographics))])))
+
+(defn- ^boolean valid-sites?
+  [project]
+  (let [sites (get-in project [:config :sites])]
+    (every? false? [(nil? (:dataset-id project))
+                    (blank? (:capacity sites))])))
+
+(defn- ^boolean valid-coverage?
+  [project]
+  (let [coverage (get-in project [:config :coverage])]
+    (not (empty? (:filter-options coverage)))))
+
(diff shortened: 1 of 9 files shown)


src/planwise/endpoint/projects2.clj
```diff
@@ -64,6 +65,7 @@
        (if (nil? project)
          (not-found {:error "Project not found"})
          (do
+           (assert (s/valid? ::project-start-model/project-starting project) "Invalid starting project")
```
ggiraldez: This will throw and generate a 500, but this should actually be a 400 since it's a bad request from the client.

src/planwise/model/starting_project.cljc
```diff
@@ -0,0 +1,36 @@
...
+(s/def ::unit-name (comp not blank?))
+(s/def ::demographics (s/keys :req-un [::unit-name ::target]))
+
+;; Sites
+(s/def ::dataset-id number?)
+(s/def ::capacity number?)
+(s/def ::sites (s/keys :req-un [::capacity]))
+
+;; Coverage
+(s/def ::filter-options map?)
```
ggiraldez: Even though we cannot be precise with the spec here since we don't have the dataset's coverage algorithm, we may still improve this using `s/or`.

src/planwise/component/projects2.clj
```diff
@@ -49,8 +49,8 @@
 
 (defn start-project
   [store project-id]
-  (db-start-project! (get-db store) {:id project-id})
   (let [project (get-project store project-id)]
+    (db-start-project! (get-db store) {:id project-id})
```
ggiraldez: Any reason for this change? This may in fact make the later call to `create-initial-scenario` with an outdated version of the project.

src/planwise/client/projects2/components/settings.cljs
```diff
@@ -15,7 +15,9 @@
             [planwise.client.ui.common :as ui]
             [planwise.client.ui.filter-select :as filter-select]
             [planwise.client.ui.rmwc :as m]
-            [planwise.client.utils :as utils]))
+            [planwise.client.utils :as utils]
+            [planwise.model.starting-project :as project-start-model]
```
ggiraldez: As long as the specs are loaded, you don't actually need this require here. The spec registry is global, so you can refer the spec later with the fully qualified namespace.

PR #400, opened 2018-05-15
Clear current scenario when scenario view component will unmount

### LINKED ISSUE #394: Browser keeps reloading a scenario even after leaving the page
Opening a `pending` scenario results in an interval/timeout being set to periodically refresh the scenario contents. This happens even after navigating away into the scenarios list, or even navigating to a different project.

ftarulla:
Fix #394

### DIFF
--- a/src/planwise/client/scenarios/handlers.cljs
+++ b/src/planwise/client/scenarios/handlers.cljs
@@ -54,12 +54,11 @@
  (fn [{:keys [db]} [_ scenario]]
    (let [current-scenario (get-in db [:scenarios :current-scenario])
          should-update    (= (:id current-scenario) (:id scenario))]
-     (merge (cond
-              should-update {:db (-> db
-                                     (assoc-in [:scenarios :current-scenario]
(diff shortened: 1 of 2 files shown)


src/planwise/client/scenarios/handlers.cljs
```diff
@@ -54,12 +54,11 @@
...
-              should-update {:db (-> db
-                                     (assoc-in [:scenarios :current-scenario]
-                                               (merge current-scenario (select-keys scenario demand-fields))))}
-              :else {})
-            (dispatch-track-demand-information-if-needed scenario)))))
+     (if should-update
+       (merge {:db (-> db (assoc-in [:scenarios :current-scenario]
+                                    (merge current-scenario (select-keys scenario demand-fields))))}
+              (dispatch-track-demand-information-if-needed scenario))
+       {}))))
```
ggiraldez: No need to return an empty map if the handler shouldn't produce an effect. You can use `when` instead of `if`.

Issue #408, opened 2018-05-24
Prevent creating sites in initial scenario


solsarratea:
Suggestion: block "go-back" button when create-site-dialog is opened

PR #409, opened 2018-05-24
Fix/create site dialog

solsarratea:
Continues in #408

### DIFF
--- a/src/planwise/client/components/common2.cljs
+++ b/src/planwise/client/components/common2.cljs
@@ -33,16 +33,19 @@
    [:p "Loading..."]])
 
 (defn text-field
-  [{:keys [label value] :as props}]
-  (let [focus  (r/atom false)
-        id     (str (random-uuid))]
-    (fn [{:keys [label value] :as props}]
-      (let [props (dissoc props :labels)]
-        [:div.mdc-text-field.mdc-text-field--upgraded {:class (when @focus "mdc-text-field--focused")}
-         [:input.mdc-text-field__input (merge props {:id id
-                                                     :on-focus #(reset! focus true)
-                                                     :on-blur  #(reset! focus false)})]
-         [:label.mdc-floating-label {:for id
-                                     :class (when (or (not (blank? (str value))) @focus) "mdc-floating-label--float-above")}
-          label]
-         [:div.mdc-line-ripple {:class (when @focus "mdc-line-ripple--active")}]]))))
+  ([props-input]
+   (text-field props-input nil))
+  ([{:keys [label value] :as props-input} custom-text-prop]
+   (let [focus  (r/atom false)
+         id     (str (random-uuid))]
+     (fn [{:keys [class label value] :as props-input} custom-text-prop]
+       (let [props (dissoc props-input :label)]
+         [:div.mdc-text-field.mdc-text-field--upgraded {:class (when @focus (str "mdc-text-field--focused" custom-text-prop))}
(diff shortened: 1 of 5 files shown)


src/planwise/client/components/common2.cljs
```diff
@@ -33,16 +33,19 @@
...
-         [:div.mdc-line-ripple {:class (when @focus "mdc-line-ripple--active")}]]))))
+  ([props-input]
+   (text-field props-input nil))
+  ([{:keys [label value] :as props-input} custom-text-prop]
+   (let [focus  (r/atom false)
+         id     (str (random-uuid))]
+     (fn [{:keys [class label value] :as props-input} custom-text-prop]
+       (let [props (dissoc props-input :label)]
+         [:div.mdc-text-field.mdc-text-field--upgraded {:class (when @focus (str "mdc-text-field--focused" custom-text-prop))}
+          [:input.mdc-text-field__input (merge props-input  {:id id
```
ggiraldez: I think you meant `props` here instead of `props-input`, which is computed above.

src/planwise/client/components/common2.cljs
```diff
@@ -33,16 +33,19 @@
...
-         [:input.mdc-text-field__input (merge props {:id id
-                                                     :on-focus #(reset! focus true)
-                                                     :on-blur  #(reset! focus false)})]
-         [:label.mdc-floating-label {:for id
-                                     :class (when (or (not (blank? (str value))) @focus) "mdc-floating-label--float-above")}
-          label]
-         [:div.mdc-line-ripple {:class (when @focus "mdc-line-ripple--active")}]]))))
+  ([props-input]
+   (text-field props-input nil))
+  ([{:keys [label value] :as props-input} custom-text-prop]
```
ggiraldez: I would prefer to include the extra argument inside the props map. Maybe something like `focus-extra-class` or something along those lines?

src/planwise/client/scenarios/edit.cljs
```diff
@@ -26,31 +26,29 @@
...
-  [{:keys [investment capacity]}]
-  [:div
-   [:h2 "Investment"]
-   [input {:value (or investment "")
-           :onchange-path [:changeset-dialog :investment]}]
-   [:h2 "Capacity"]
-   [input {:value (or capacity "")
-           :onchange-path [:changeset-dialog :capacity]}]])
+  [{:keys [investment capacity]} scenario-investment project-budget]
+  (let [custom-text-prop (when (< project-budget (+ scenario-investment investment)) " invalid-input")]
```
ggiraldez: Intent here is to compute if the budget is valid or not, which is kind of lost in translation. So, I'd compute that (as a boolean) and use it later to output the "extra class".

src/planwise/client/scenarios/edit.cljs
```diff
@@ -26,31 +26,29 @@
...
 (defn changeset-dialog-content
-  [{:keys [investment capacity]}]
-  [:div
-   [:h2 "Investment"]
-   [input {:value (or investment "")
-           :onchange-path [:changeset-dialog :investment]}]
-   [:h2 "Capacity"]
-   [input {:value (or capacity "")
-           :onchange-path [:changeset-dialog :capacity]}]])
+  [{:keys [investment capacity]} scenario-investment project-budget]
```
ggiraldez: It's better to receive a `remaining-budget` or `available-budget` here. That's one less parameter and clearer what the purpose of the additional data is.

PR #429, opened 2018-06-05
Population tables

### LINKED ISSUE #410: Add `sources` table and migrate existing `population_sources`

Sources refer to the individuals/sites that after being filtered, constitute the demand for the project.

- Migrate existing tables and create `sources` to contain the discrete consumer locations
- Existing population sources should be converted into raster type source sets, with unit "people"

Roughly:

source_set:
---
id
name
type: raster | points
unit
raster_file

sources:
---
id
set_id
name
type
lat, lon
quantity

ftarulla:
Fix #410


resources/migrations/055-update-population-source-id-to-projects2.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE projects2
+  DROP COLUMN IF EXISTS "population-source-id",
+  ADD COLUMN "population-source-id" BIGINT REFERENCES sources_set(id) DEFAULT NULL;
```
ggiraldez: This should be `source-set-id` right?

PR #430, opened 2018-06-08
Feature/display providers in scenario

solsarratea:


src/planwise/client/scenarios/views.cljs
```diff
@@ -16,16 +16,23 @@
             [planwise.client.utils :as utils]
             [planwise.client.ui.rmwc :as m]))
 
+(defn- show-provider
+  [{{:keys [action name capacity investment]} :elem :as provider}]
+  (if (nil? action)
+    (str "<b>" name "</b><br> Capacity: " capacity)
```
ggiraldez: Since `name` is user provided, it should be properly escaped to avoid Javascript injection attacts. You can use `planwise.clients.utils/escape-html`.

src/planwise/client/scenarios/views.cljs
```diff
@@ -16,16 +16,23 @@
...
         index    (subscribe [:scenarios/changeset-index])
         position (r/atom mapping/map-preview-position)
         zoom     (r/atom 3)
         add-point (fn [lat lon] (dispatch [:scenarios/create-provider {:lat lat
                                                                        :lon lon}]))]
-    (fn [{:keys [bbox]} {:keys [changeset raster]}]
-      (let [indexed-changeset     (map (fn [elem] {:elem elem :index (.indexOf changeset elem)}) changeset)
+    (fn [{:keys [bbox]} {:keys [changeset providers raster] :as scenario}]
+      (let [changeset   (into providers changeset)
+            indexed-changeset     (map (fn [elem] {:elem elem :index (.indexOf changeset elem)}) changeset)
```
ggiraldez: Pretty sure `.indexOf` doesn't work here, since `changeset` should be a Clojurescript structure and not a Javascript array. You can use `map-indexed` to achieve similar results.

src/planwise/component/scenarios.clj
```diff
@@ -44,6 +46,26 @@
...
 
+(defn- get-initial-providers
+  [store provider-set-id version filter-options]
+  (let [providers (providers-set/get-providers-with-coverage-in-region
+                   (:providers-set store) provider-set-id version filter-options)
+        select-fn (fn [{:keys [id name capacity lat lon]}]
+                    {:initial true :provider-id (str id)
+                     :name name    :capacity capacity
+                     :location {:lat lat :lon lon}})]
+    (seq (map select-fn providers))))
```
ggiraldez: `map` already returns a lazy sequence, so `seq` is not necessary here.

test/planwise/component/scenarios_test.clj
```diff
@@ -108,13 +108,14 @@
...
+    [{:action "create-provider" :provider-id "new.1" :investment 1000 :capacity 50 :location {:lat 0 :lon 0}}]
     [{:action "create-provider" :provider-id "new.1" :investment 10000 :capacity 50 :location {:lat 0 :lon 0}}]))
 
 (deftest invalid-changeset
   (are [x] (not (s/valid? ::model/change-set x))
-    [{:action "unknown-action" :provider-id "new.1" :investment 10000 :capacity 50 :location {:lat 0 :lon 0}}]
+    [{:action "unknown-action"  :provider-id "new.1" :investment 10000 :capacity 50 :location {:lat 0 :lon 0}}]
     [{:action "create-provider" :provider-id "new.1" :investment nil :capacity nil}]
-    [{:action "create-provider" :provider-id "new.1" :investment "" :capacity ""}]
+    [{:action "create-provider" :provider-id "nmew.1" :investment "" :capacity ""}]
```
ggiraldez: Typo?

src/planwise/model/scenarios.clj
```diff
@@ -6,10 +6,15 @@
...
-  (s/keys :req-un [:create-provider/action ::investment ::capacity ::provider-id ::location]))
+(s/def ::create-provider
+  (s/keys :req-un [::action ::investment ::capacity ::provider-id ::location]))
+
+(s/def ::initial-provider
+  (s/keys :req-un [::initial ::provider-id ::capacity ::location]))
 
 (s/def ::change-set
-  (s/coll-of ::create-provider-change))
+  (s/coll-of (s/or :create-provider ::create-provider
```
ggiraldez: I think this is a leftover from a previous implementation. AFAICS, you are not putting initial providers in the `changeset` (which is good btw), but in a separate `:providers` key.

src/planwise/component/scenarios.clj
```diff
@@ -44,6 +46,26 @@
   (-> (db-find-scenario (get-db store) {:id scenario-id})
       (update :changeset edn/read-string)))
 
+(defn- get-initial-providers
+  [store provider-set-id version filter-options]
+  (let [providers (providers-set/get-providers-with-coverage-in-region
+                   (:providers-set store) provider-set-id version filter-options)
+        select-fn (fn [{:keys [id name capacity lat lon]}]
+                    {:initial true :provider-id (str id)
```
ggiraldez: I don't think you need the `:initial` key here. Also, why convert the `id` into a string?

PR #432, opened 2018-06-13
Feature/416 export csv from scenario

### LINKED ISSUE #416: Export CSV from any scenario
Story: **Users can export satisfied and unsatisfied demand**

Related to #415. Allow the user to export a CSV file with the scenario providers with both the demand they are satisfying and the unsatisfied demand under their coverage area.

solsarratea:
Closes #416

### DIFF
--- a/src/planwise/component/scenarios.clj
+++ b/src/planwise/component/scenarios.clj
@@ -42,11 +42,9 @@
 
 (defn- map->csv
   [file-name coll]
-  {:coll coll}
   (let [fields (mapv name (keys (first coll)))
         data   (cons fields (mapv vals coll))]
-    (with-open [out-file (io/writer file-name)]
-      (csv/write-csv out-file data))))
+    (with-out-str (csv/write-csv *out* data))))
 
 ;; ----------------------------------------------------------------------
 ;; Service definition
@@ -263,7 +261,9 @@
   (reset-scenarios [store project-id]
     (reset-scenarios store project-id))
   (get-scenario-for-project [store scenario project]
-    (get-scenario-for-project store scenario project)))
+    (get-scenario-for-project store scenario project))
+  (export-providers-data [store scenario-id]
+    (export-providers-data store scenario-id)))
 
 (defmethod ig/init-key :planwise.component/scenarios
   [_ config]
--- a/src/planwise/component/scenarios.clj
+++ b/src/planwise/component/scenarios.clj
@@ -41,9 +41,10 @@
         (format "Create %d %s. Increase overall capacity in %d." providers u capacity))))
 
 (defn- map->csv
-  [file-name coll]
-  (let [fields (mapv name (keys (first coll)))
-        data   (cons fields (mapv vals coll))]
+  [coll]
+  (let [fields (keys (first coll))
+        rows   (map #(map % fields) coll)
+        data   (cons (mapv name fields) rows)]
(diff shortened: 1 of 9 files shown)


src/planwise/component/scenarios.clj
```diff
@@ -38,6 +40,13 @@
     (if (zero? providers) ""
         (format "Create %d %s. Increase overall capacity in %d." providers u capacity))))
 
+(defn- map->csv
+  [file-name coll]
+  {:coll coll}
+  (let [fields (mapv name (keys (first coll)))
+        data   (cons fields (mapv vals coll))]
```
ggiraldez: While it is not likely to happen if the maps of the collection are created by the same code, there are no guarantees that the order of the keys of the first map will match the order in which `vals` returns the values. Ie. the order for `keys` and `vals` *for the same map* is guaranteed to be the same, but the data is obtained from multiple maps.

Eg.

```clojure
(def a {:id 1 :name "foo"})
(def b {:name "bar" :id 2})
(keys a) ;;=> (:id :name)
(vals a) ;;=> (1 "foo")
(vals b) ;;=> ("bar" 2)
```

src/planwise/component/scenarios.clj
```diff
@@ -206,6 +215,32 @@
            (cons name)
            (util-str/next-name))))
 
+(defn- changeset-to-export
+  [changeset]
+  (mapv (fn [{:keys [provider-id action capacity location]}]
+          {:provider-id provider-id      :type action
+           :name "" :lat (:lat location) :lon (:lon location)
+           :capacity capacity            :tags ""}) changeset))
```
ggiraldez: This formatting is weird. Please put a single key value on each line.

src/planwise/component/scenarios.clj
```diff
@@ -42,11 +42,9 @@
 
 (defn- map->csv
   [file-name coll]
```
ggiraldez: `file-name` is no longer used.

src/planwise/endpoint/scenarios.clj
```diff
@@ -30,17 +30,18 @@
...
            project-id (:project-id (scenarios/get-scenario service id))
            project    (filter-owned-by (projects2/get-project projects2 project-id) user-id)]
-       (if (or (nil? project) (nil? scenario))
-         (not-found {:error "Scenario not found"})
-         (do
-           (scenarios/update-scenario service project scenario)
-           (response (dissoc (scenarios/get-scenario service id) scenario :updated-at))))))
+       (if (or (nil? project) (nil? scenario)) (not-found {:error "Scenario not found"})
+           (do (scenarios/update-scenario service project scenario)
+               (response (dissoc (scenarios/get-scenario service id) scenario :updated-at))))))
```
ggiraldez: You should keep the `then` expression in a separate line.

src/planwise/component/scenarios.clj
ggiraldez: This is still relevant after the changes in 2cfb7e2

src/planwise/endpoint/scenarios.clj
ggiraldez: Ditto.

ggiraldez:
Please rebase and resolve. Optionally change the CSV mapping code as per the comment, but I'll leave you to decide.

Issue #413, opened 2018-05-24
Import consumers/sources CSV file


ggiraldez:
Story: **Users can import CSV with demand**

When creating a new source/consumer set, the user can provide a CSV file with the locations and number of potential consumers.

PR #441, opened 2018-06-21
Sources owner

### LINKED ISSUE #440: User created sources should have an owner
Add a new field to the table and filter the returned sources by the current user. System wide sources would have the owner field as NULL.

ftarulla:
Fix #440


ggiraldez:
Oops! Spoke too soon. You are missing setting the ownership when creating a new source. Please add that and we're good to go!

ggiraldez:
Never mind. CSV import is done in #413.

PR #442, opened 2018-06-26
Import sources from csv

### LINKED ISSUE #413: Import consumers/sources CSV file

ftarulla:
Fix #413

### DIFF
--- a/src/planwise/component/sources.clj
+++ b/src/planwise/component/sources.clj
@@ -2,7 +2,10 @@
   (:require [planwise.boundary.sources :as boundary]
             [integrant.core :as ig]
             [taoensso.timbre :as timbre]
-            [hugsql.core :as hugsql]))
+            [clojure.java.io :as io]
+            [clojure.data.csv :as csv]
+            [hugsql.core :as hugsql]
+            [clojure.java.jdbc :as jdbc]))
 
 ;; ----------------------------------------------------------------------
 ;; Auxiliary and utility functions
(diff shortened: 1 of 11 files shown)


src/planwise/component/sources.clj
```diff
@@ -13,15 +16,69 @@
...
 
+(defn create-source-set
+  [store name unit owner-id]
+  (db-create-source-set! (get-db store) {:name name
+                                         :unit unit
+                                         :owner-id owner-id}))
+
+(defn create-source
+  [store source]
+  (println source)
```
ggiraldez: Remove the `println`

src/planwise/component/sources.clj
```diff
@@ -13,15 +16,69 @@
...
+  (jdbc/with-db-transaction [tx (get-db store)]
+    (let [tx-store                  (assoc-in store [:db :spec] tx)
+          tx-create-source-set      (fn [name unit]
+                                      (let [result (create-source-set tx-store name unit owner-id)]
+                                        (:id result)))
+          tx-import-sources-from    (fn [set-id csv-file] (import-sources tx-store set-id csv-file))
+          tx-get-created-source-set (fn [set-id] (get-source-set tx-store set-id owner-id))]
+      (-> (tx-create-source-set name unit)
+          (tx-import-sources-from csv-file)
+          (tx-get-created-source-set)))))
```
ggiraldez: This is a bit convoluted. I think a more explicit implementation would read much better. Eg.

```clojure
(let [tx-store (assoc-in store [:db :spec] tx)
      source-set (create-source-set tx-store unit name owner-id)
      source-set-id (:id source-set)]
  (import-sources tx-store source-set-id csv-file)
  (get-source-set tx-store source-set-id owner-id))
```

You can also rename `import-sources!` and `create-source-set!` to make the side effect more visible.

PR #444, opened 2018-07-13
Scenario computation discrete locations

### LINKED ISSUE #411: Change scenario computation to support discrete locations

For projects with selected source of type "points". Instead of processing a raster file, the scenario result should be a table with the same number of filtered consumers and each demand quantity reduced by the project providers (if under the coverage area of the provider).

ftarulla:
Fix #411


src/planwise/component/engine.clj
```diff
@@ -99,19 +101,97 @@
...
+                                      id-sources-under-coverage (set (map :id (fn-sources-under provider)))
+                                      sources-under-coverage    (fn-select-by-id sources id-sources-under-coverage) ; updated sources under coverage
+                                      total-demand              (sum-up sources-under-coverage :quantity)]
+                                  (assoc provider :unsatisfied total-demand)))
+                              (:providers result-step1))]
+
+    (let [updated-sources           (map (fn [source] ; persist only id and quantity
+                                           {:id (:id source)
+                                            :quantity (:quantity source)})
+                                         (:sources result-step1))
```
ggiraldez: This can also be done with `(map #(select-keys % [:id :quantity]) (:sources result-step1))`

src/planwise/client/sources/handlers.cljs
```diff
@@ -44,13 +44,14 @@
...
  :sources.new/created
  in-sources
- (fn [db [_ updated-sources-list]]
+ (fn [db [_ created-source]]
    (rf/dispatch [:modal/hide])
-   (dissoc db :new)
-   (update db :list #(asdf/reset! % updated-sources-list))))
+   (-> db
+       (dissoc :new)
+       (update :list asdf/swap! into [created-source]))))
```
ggiraldez: You can use `conj` here and avoid creating the single element vector. NB. it's exactly what `into` does https://github.com/clojure/clojure/blob/clojure-1.9.0/src/clj/clojure/core.clj#L6817

src/planwise/component/engine.clj
```diff
@@ -99,19 +101,97 @@
...
-       :covered-demand   (- source-demand initial-demand)
+       :pending-demand   scenario-demand
+       :covered-demand   (- source-demand scenario-demand)
        :demand-quartiles quartiles
        :providers-data   (mapv (fn [[a b]] (merge a b)) (map vector processed-providers update-providers))})))
 
-(defn compute-scenario
+(defn sum-up
+  [coll k]
+  (reduce (fn [acum elem] (+ acum (k elem))) 0 coll))
```
ggiraldez: `mapsum` or `sum-map` would be more descriptive names. Also, `k` can be called `f` to denote that's it's a generic function.

Another way of writing this function would be: `(apply + (map k coll))`

src/planwise/client/sources/handlers.cljs
```diff
@@ -48,7 +48,7 @@
    (rf/dispatch [:modal/hide])
    (-> db
        (dissoc :new)
-       (update :list asdf/swap! into [created-source]))))
+       (update :list #(asdf/swap! % conj created-source)))))
```
ggiraldez: The anonymous function is not needed here, since the ordering of the arguments does not change.

PR #448, opened 2018-07-19
Render point based sources

### LINKED ISSUE #447: Show point based sources in scenario maps
For projects which use a point-type source set, allow the user to see the sources in the map with their respective quantity. Color the map points using quartiles, as in the case of a raster source.

ftarulla:
Fix #447


src/planwise/component/scenarios.clj
```diff
@@ -69,14 +70,23 @@
...
 
 (defn get-scenario-for-project
-  [store scenario {:keys [provider-set-id provider-set-version config] :as project}]
+  [store scenario {:keys [provider-set-id provider-set-version config source-set-id] :as project}]
   (let [filter-options (-> (select-keys project [:region-id :coverage-algorithm])
                            (assoc :tags (get-in config [:providers :tags])
                                   :coverage-options (get-in config [:coverage :filter-options])))
-        initial-providers  (get-initial-providers store provider-set-id provider-set-version filter-options)]
+        initial-providers  (get-initial-providers store provider-set-id provider-set-version filter-options)
+        sources-data       (edn/read-string (:sources-data scenario))
```
ggiraldez: Serialization format is an implementation detail which should be moved into a lower "layer". Eg. deserialize in `get-scenario`.

src/planwise/client/scenarios/views.cljs
```diff
@@ -22,17 +22,30 @@
...
-                                                                       :lon lon}]))]
+                                                                       :lon lon}]))
+        use-providers-clustering false
+        providers-type-layer     (if use-providers-clustering :cluster-layer :point-layer)]
     (fn [{:keys [bbox]} {:keys [changeset providers raster] :as scenario}]
-      (let [providers           (into providers changeset)
-            indexed-providers   (map-indexed (fn [idx elem] {:elem elem :index idx}) providers)
+      (let [providers             (into providers changeset)
+            indexed-providers     (to-indexed-map providers)
+            indexed-sources       (to-indexed-map (:sources scenario))
```
ggiraldez: You don't need to index sources. In fact, indexing providers for identification later when editing is wrong as well, although pre-existing.

src/planwise/endpoint/home.clj
```diff
@@ -103,6 +103,10 @@
       (include-js "/js/leaflet.bboxloader.js")
       (include-js "/js/leaflet.legend.js")
       (include-js "/js/main.js")
+      (include-js "/js/leaflet.markercluster.js")
+      (include-css "/css/MarkerCluster.css")
+      (include-css "/css/MarkerCluster.Default.css")
+      (include-css "/css/MarkerCluster.Custom.css")
```
ggiraldez: CSS should be included from the header, along with the other CSS files. And we should probably integrate these files into the SASS compilation.

PR #452, opened 2018-07-24
Fix/408 create sites

### LINKED ISSUE #408: Prevent creating sites in initial scenario

solsarratea:
Closes #408

### DIFF
--- a/src/planwise/client/scenarios/handlers.cljs
+++ b/src/planwise/client/scenarios/handlers.cljs
@@ -92,14 +92,8 @@
  :scenarios/clear-current-scenario
  in-scenarios
  (fn [db [_]]
-   (assoc db :current-scenario nil)))
-
-(rf/reg-event-db
- :scenarios/clear-state
- in-scenarios
- (fn [db [_]]
-   (assoc db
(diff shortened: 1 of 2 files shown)


src/planwise/client/scenarios/handlers.cljs
```diff
@@ -152,10 +153,13 @@
  :scenarios/open-changeset-dialog
  in-scenarios
  (fn [db [_ changeset-index]]
-   (assoc db
-          :view-state        :changeset-dialog
-          :view-state-params {:changeset-index changeset-index}
-          :changeset-dialog  (get-in db [:current-scenario :changeset changeset-index]))))
+   (let [last-change (:changeset-dialog db)
+         provider    (get-in db [:current-scenario :changeset changeset-index])
+         condition   (reduce = (map :provider-id [provider last-change]))]
```
ggiraldez: This looks like a complicated way of saying `(= (:provider-id provider) (:provider-id last-change))`. Also, what is the condition being tested here? And why changing the contents of `changeset-dialog` depend on it?

PR #453, opened 2018-07-26
Feature/419 algorithm for new provider

### LINKED ISSUE #419: Design algorithm to find best location for a new provider
Story: **Planwise calculates best location for a given intervention**

Given a geographic demand distribution (either a raster file, or a set of discrete points), the algorithm should find the best location for creating a new provider to satisfy demand given a coverage algorithm configured with some parameters. "Best" should be interpreted as the location which yields coverage to the greatest amount of demand. There might be more than one location which satisfies that criteria, and so the algorithm may respond with a list of best candidates.

solsarratea:
Closes #419 

Improvements may be suggested :)
Results of testing:


src/planwise/engine/raster.clj
```diff
@@ -65,6 +65,23 @@
...
+        raster-count (.GetRasterCount dataset)
+        band      (.GetRasterBand dataset 1)
+        data-type (.GetRasterDataType band)
+        xsize     (.GetXSize band)
+        ysize     (.GetYSize band)
+        nodata    (read-nodata-value band)
+        data      (read-band-data band 0 0 xsize ysize)]
+    (->Raster projection geotransform xsize ysize data-type nodata data)))
+
+(defn update-raster
```
ggiraldez: Naming is a bit confusing. Probably move this into a new variant of `create-raster`, since this really is creating a new raster.

PR #472, opened 2018-08-28
Disable "Start" when settings are invalid

### LINKED ISSUE #469: Disable "Start" project when settings are incomplete/invalid
This may happen when changing coverage algorithm, which in turn changes the coverage criteria options. The UI is left inconsistent with the underlying model. Eg.
- select a provider set using the walking friction coverage algo
- select a walking time
- select a different provider set using the driving friction coverage
The criteria combo for driving time is rendered with unselected options, but the underlying model still has the previously selected option for "walking time". Thus the project is evaluated as valid and the start button is enabled, but then later the coverage will fail because of invalid criteria options.

solsarratea:
Validate filtering-options when changing provider-set
Fixes #469

### DIFF
--- a/src/planwise/client/projects2/handlers.cljs
+++ b/src/planwise/client/projects2/handlers.cljs
@@ -127,14 +127,15 @@
 (rf/reg-event-fx
  :projects2/update-current-project-from-server
  in-projects2
- (fn [{:keys [db]} [_ project]]
+ (fn [{:keys [db]} [_ {:keys [config] :as project}]]
    (let [current-project (:current-project db)]
      (if (= (:id project) (:id current-project))
         ;; keep current values of current-project except the once that could be updated from server
         ;; to prevent replacing data that have not been saved in server yet
        (let [updated-project (-> current-project
                                  (assoc :providers (:providers project))
-                                 (assoc :coverage-algorithm (:coverage-algorithm project)))]
+                                 (assoc :coverage-algorithm (:coverage-algorithm project))
+                                 (assoc-in [:config :coverage :filter-options] (get-in config [:coverage :filter-options])))]
          {:dispatch [:projects2/save-project-data updated-project]})))))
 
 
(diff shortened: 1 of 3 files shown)


src/planwise/client/projects2/handlers.cljs
```diff
@@ -127,14 +127,15 @@
...
+ (fn [{:keys [db]} [_ {:keys [config] :as project}]]
    (let [current-project (:current-project db)]
      (if (= (:id project) (:id current-project))
         ;; keep current values of current-project except the once that could be updated from server
         ;; to prevent replacing data that have not been saved in server yet
        (let [updated-project (-> current-project
                                  (assoc :providers (:providers project))
-                                 (assoc :coverage-algorithm (:coverage-algorithm project)))]
+                                 (assoc :coverage-algorithm (:coverage-algorithm project))
+                                 (assoc-in [:config :coverage :filter-options] (get-in config [:coverage :filter-options])))]
```
ggiraldez: This is fine, but it'd be better if updating `:filter-options` would only happen if there's a change in the incoming `:coverage-algorithm`.

src/planwise/component/projects2.clj
```diff
@@ -32,10 +34,17 @@
...
-  (db-update-project (get-db store) (-> project
-                                        (update :config pr-str)
-                                        (dissoc :state))))
+  [store {:keys [config provider-set-id] :as project}]
+  (let [algorithm (-> (providers-set/get-provider-set (:providers-set store) provider-set-id)
+                      :coverage-algorithm
+                      keyword)
+        valid-criteria? (s/valid? ::coverage/criteria
+                                  (assoc (get-in project [:config :coverage :filter-options])
+                                         :algorithm algorithm))
```
ggiraldez: Can you add this validation check to the `start-project` function? That way, even if something's wrong in the client side, we still don't start the project with an invalid coverage configuration. Currently it's making a spec check, but since the coverage algorithm is in the provider set and not in the project, it can't verify that the coverage options are actually correct.

solsarratea: Shouldn't specs also validate that criteria has only one filter option?  
Maybe using  :max-count 1 https://github.com/clojure/spec.alpha/blob/b2b5db433cad9f1ad56bd28f58 4136cef2bae73e/src/main/clojure/clojure/spec/alpha.clj#L545

ggiraldez: No, not really, because coverage algorithms may require more than one parameter.

PR #476, opened 2018-08-29
Requests cant be canceled

### LINKED ISSUE #456: Requests can't be canceled
Using reframe 0.10.5. 

When using the effects `:db` `:api`, the order is important when the effect associated with `:api` changes the DB (although the documentation says the effects run simultaneously)
Effects associated with `:api` **shouldn't modify the DB directly**!

solsarratea:
Closes #456

### DIFF
--- a/src/planwise/client/effects.cljs
+++ b/src/planwise/client/effects.cljs
@@ -102,8 +102,7 @@
                       (dissoc :key)
                       request->xhrio-options
                       ajax/ajax-request)]
-        ;FIXME: Issue #456
-        (when key (swap! rf-db/app-db assoc-in key xhrio))))))
+        (when key (swap! rf-db/app-db assoc-in [key] xhrio))))))
 
 
 (rf/reg-fx :api api-effect)
@@ -112,7 +111,7 @@
  :api-abort
  (fn [key]
    (some-> (get-in @rf-db/app-db key) ajax.protocols/-abort)
-   (swap! rf-db/app-db assoc-in key nil)))
+   (swap! rf-db/app-db dissoc key)))
 
 (defn make-api-request
   "Allows manually triggering an API request. Use :on-success-cb and
--- a/src/planwise/client/effects.cljs
+++ b/src/planwise/client/effects.cljs
@@ -112,6 +112,7 @@
 (rf/reg-fx
  :api-abort
  (fn [key]
+   (some-> (key @on-request) ajax.protocols/-abort)
    (swap! on-request dissoc key)))
 
 (defn make-api-request
--- a/src/planwise/client/effects.cljs
+++ b/src/planwise/client/effects.cljs
@@ -58,9 +58,11 @@
 ;; OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 ;; SOFTWARE.
 
+(def on-request (atom {}))
+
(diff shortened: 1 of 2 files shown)


src/planwise/client/scenarios/handlers.cljs
```diff
@@ -251,28 +251,34 @@
...
+   (let [request-key (str (random-uuid))]
+     {:db  (-> db (assoc-in [:current-scenario :computing-best-locations :state] request-key)
+               (assoc :view-state :get-suggestions))
+      :api (assoc (api/suggested-providers (get-in db [:current-scenario :id]))
+                  :on-success [:scenarios/suggested-providers]
+                  :on-failure [:scenarios/no-suggested-providers]
+                  :key request-key)})))
 
-(rf/reg-event-db
+(rf/reg-event-fx
```
ggiraldez: This should be a `reg-event-db` or return the appropriate effects map.

src/planwise/client/effects.cljs
```diff
@@ -102,8 +102,7 @@
                       (dissoc :key)
                       request->xhrio-options
                       ajax/ajax-request)]
-        ;FIXME: Issue #456
-        (when key (swap! rf-db/app-db assoc-in key xhrio))))))
+        (when key (swap! rf-db/app-db assoc-in [key] xhrio))))))
```
ggiraldez: This still modifies the re-frame database directly, which we should be trying to avoid. Instead use a separate atom local to this namespace.

src/planwise/client/effects.cljs
```diff
@@ -112,7 +111,7 @@
  :api-abort
  (fn [key]
    (some-> (get-in @rf-db/app-db key) ajax.protocols/-abort)
-   (swap! rf-db/app-db assoc-in key nil)))
+   (swap! rf-db/app-db dissoc key)))
```
ggiraldez: As above, and we would also need to dissoc the key when the request finishes correctly.

src/planwise/client/scenarios/handlers.cljs
```diff
@@ -251,28 +251,34 @@
...
 (rf/reg-event-fx
  :scenarios.new-provider/fetch-suggested-locations
  in-scenarios
  (fn [{:keys [db]} [_]]
-   {:db  (-> db (assoc-in [:current-scenario :computing-best-locations :state] true)
-             (assoc :view-state :get-suggestions))
-    :api (assoc (api/suggested-providers (get-in db [:current-scenario :id]))
-                :on-success [:scenarios/suggested-providers]
-                :on-failure [:scenarios/no-suggested-providers])}))
+   (let [request-key (str (random-uuid))]
```
ggiraldez: We don't need a random id here. Since there can only be one in-flight request for suggestions, we can use some hard coded value for this key (eg. `:suggestions-request`).

src/planwise/client/effects.cljs
```diff
@@ -102,16 +104,15 @@
...
 
 
 (rf/reg-fx :api api-effect)
 
 (rf/reg-fx
  :api-abort
  (fn [key]
-   (some-> (get-in @rf-db/app-db key) ajax.protocols/-abort)
-   (swap! rf-db/app-db dissoc key)))
+   (swap! on-request dissoc key)))
```
ggiraldez: With this change you are not actually aborting the request.

Issue #473, opened 2018-08-28
Add sample csv for Sources


bcardiff:
And a basic text hint regarding the Unit field.

PR #485, opened 2018-08-30
Add fields when exporting .csv

### LINKED ISSUE #480: Add demand and capacity fields to the exported .csv
Add fields described in #479.
Also the download csv of scenario has no proper filename + extension. it's named 'csv' make it '{project-id}-{scenario-id}-{scenario-name}.csv'

solsarratea:
Closes #480 (Rebased branch onto Master)
Sources sample in #473 should be updated (DONE)

### DIFF
--- a/src/planwise/client/scenarios/views.cljs
+++ b/src/planwise/client/scenarios/views.cljs
@@ -240,10 +240,9 @@
         created-providers (subscribe [:scenarios/created-providers])
         source-demand (get-in current-project [:engine-config :source-demand])
         unit-name  (get-in current-project [:config :demographics :unit-name])
-        csv-name (str (:id current-project) "-" id "-" (:name current-scenario) ".csv")
         export-providers-button [:a {:class "mdc-fab disable-a"
                                      :id "main-action"
-                                     :href (str "/api/scenarios/" id "/" csv-name)}
+                                     :href (str "/api/scenarios/" id "/csv")}
                                  [m/Icon {:class "material-icons  center-download-icon"} "get_app"]]]
     (fn [current-project current-scenario]
       [ui/full-screen (merge (common2/nav-params)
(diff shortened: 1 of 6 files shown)


src/planwise/client/scenarios/views.cljs
```diff
@@ -233,16 +233,17 @@
...
         state      (subscribe [:scenarios/view-state])
         error      (subscribe [:scenarios/error])
         created-providers (subscribe [:scenarios/created-providers])
         source-demand (get-in current-project [:engine-config :source-demand])
         unit-name  (get-in current-project [:config :demographics :unit-name])
+        csv-name (str (:id current-project) "-" id "-" (:name current-scenario) ".csv")
         export-providers-button [:a {:class "mdc-fab disable-a"
                                      :id "main-action"
-                                     :href (str "/api/scenarios/" (:id current-scenario) "/csv")}
+                                     :href (str "/api/scenarios/" id "/" csv-name)}
```
ggiraldez: Constructing a URL from a user input is generally a bad idea, as it may lead to malformed URLs, malicious attacks and data leakage. Output file name can be set via a HTTP header in the response instead.

src/planwise/endpoint/scenarios.clj
```diff
@@ -30,7 +30,7 @@
          (not-found {:error "Scenario not found"})
          (response (scenarios/get-scenario-for-project service scenario project)))))
 
-   (GET "/:id/csv" [id :as request]
+   (GET "/:id/:name" [id :as request]
      (response (scenarios/export-providers-data service (Integer. id))))
```
ggiraldez: Set a download filename by setting the [`Content-Disposition`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) HTTP [header](http://ring-clojure.github.io/ring/ring.util.response.html#var-header).

PR #493, opened 2018-09-18
Show sources and providers with different symbols

solsarratea:
Refactor scenarios view code and add classes to .css

### DIFF
--- a/src/leaflet/core.cljs
+++ b/src/leaflet/core.cljs
@@ -67,7 +67,7 @@
       (.on marker "mouseout" #(mouseout-fn marker % point)))
     marker))
 
-(defn create-point [point {:keys [lat-fn lon-fn style-fn popup-fn mouseover-fn mouseout-fn], :or {lat-fn :lat, lon-fn :lon}, :as props}]
+(defn create-point [point {:keys [lat-fn lon-fn style-fn icon-fn popup-fn mouseover-fn mouseout-fn shape], :or {lat-fn :lat, lon-fn :lon}, :as props}]
   (let [latLng    (.latLng js/L (lat-fn point) (lon-fn point))
         attrs     (dissoc props :lat-fn :lon-fn :popup-fn)
         clickable (boolean popup-fn)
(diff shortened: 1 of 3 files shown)


src/leaflet/core.cljs
```diff
@@ -67,15 +67,15 @@
       (.on marker "mouseout" #(mouseout-fn marker % point)))
     marker))
 
-(defn create-point [point {:keys [lat-fn lon-fn style-fn popup-fn mouseover-fn mouseout-fn], :or {lat-fn :lat, lon-fn :lon}, :as props}]
+(defn create-point [point {:keys [lat-fn lon-fn style-fn icon-fn popup-fn mouseover-fn mouseout-fn shape], :or {lat-fn :lat, lon-fn :lon}, :as props}]
```
ggiraldez: Looks like `shape` parameter is a leftover.

ggiraldez: I think it'd be preferable to use a `:marker-layer` instead of changing the elements for the `:point-layer`. Marker layer already has an `icon-fn` parameter even.

src/planwise/client/scenarios/views.cljs
```diff
@@ -99,100 +99,87 @@
         use-providers-clustering false
         providers-layer-type     (if use-providers-clustering :cluster-layer :point-layer)]
     (fn [{:keys [bbox]} {:keys [changeset raster sources-data] :as scenario} state error]
-      (let [indexed-providers     (to-indexed-map @all-providers)
-            indexed-sources       (to-indexed-map sources-data)
-            pending-demand-raster raster]
+      (let [indexed-providers       (to-indexed-map @all-providers)
+            indexed-sources         (to-indexed-map sources-data)
+            pending-demand-raster   raster
+            sources-layer           [:point-layer {:points indexed-sources
```
ggiraldez: Here, use `:marker-layer` instead.

src/planwise/client/scenarios/views.cljs
```diff
@@ -99,100 +99,87 @@
         use-providers-clustering false
         providers-layer-type     (if use-providers-clustering :cluster-layer :point-layer)]
```
ggiraldez: Ditto for providers layer.

PR #496, opened 2018-09-21
Fix showing changeset

solsarratea:
New providers name are created once the action is created, does not depend on the all-providers subscription.
New providers are added at the end of changeset, and edited changes do not modify the order of changeset.
Subscription for showing actions in side-panel have same order as changeset.

### DIFF
--- a/src/planwise/client/scenarios/handlers.cljs
+++ b/src/planwise/client/scenarios/handlers.cljs
@@ -178,10 +178,14 @@
  (fn [{:keys [db]} [_]]
    (let [current-scenario  (get-in db [:current-scenario])
          updated-provider  (get-in db [:changeset-dialog])
+         new-change?       (empty? (filter #(= (:id %) (:id updated-provider)) (:changeset current-scenario)))
          updated-scenario  (update current-scenario
                                    :changeset
-                                   (fn [c] (conj (utils/remove-by-id c (:id updated-provider))
-                                                 (:change updated-provider))))]
+                                   (fn [c]
+                                    (if new-change?
+                                     (conj (vec c) (:change updated-provider))
+                                     (utils/replace-by-id c (:change updated-provider))
+                                     )))]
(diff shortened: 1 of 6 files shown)


src/planwise/client/scenarios/handlers.cljs
```diff
@@ -142,20 +142,27 @@
                 (assoc-in [:current-scenario :name] name)
                 (assoc-in [:view-state] :current-scenario))})))
 
+(defn new-provider-name
+  [changeset]
+  (let [new-providers (filter #(= (:action %) "create-provider") changeset)]
+    (if (empty? new-providers)
+      "New provider 0"
+      (let [vals (mapv (fn [p] (->> (:name p) (re-find #"\d+") js/parseInt)) new-providers)]
```
ggiraldez: This is a nice piece of code! But, it "fails" if the name doesn't contain digits (which shouldn't happen, I know), since `(js/parseInt nil)` is NaN. You can fix replacing `js/parseInt` by `int` (which in Clojurescript does what you want, but not in Clojure) or `(fnil js/parseInt 0)`. Ie:
```clojure
(mapv (fn [p] (->> (:name p) (re-find #"\d+") int)) new-providers)
```
or
```clojure
(mapv (fn [p] (->> (:name p) (re-find #"\d+") ((fnil js/parseInt 0)))) new-providers)
```

src/planwise/client/scenarios/handlers.cljs
```diff
@@ -171,10 +178,13 @@
  (fn [{:keys [db]} [_]]
    (let [current-scenario  (get-in db [:current-scenario])
          updated-provider  (get-in db [:changeset-dialog])
+         new-change?       (empty? (filter #(= (:id %) (:id updated-provider)) (:changeset current-scenario)))
```
ggiraldez: You can use `utils/find-by-id` here.

PR #497, opened 2018-09-25
Delete scenario

solsarratea:
Add settings menu for either renaming or deleting scenario in current view


src/planwise/client/scenarios/handlers.cljs
```diff
@@ -325,3 +326,44 @@
...
+          :view-state :delete-scenario)))
+
+(rf/reg-event-db
+ :scenarios/open-delete-dialog
+ in-scenarios
+ (fn [db [_]]
+   (assoc db
+          :view-state :delete-scenario)))
+
+;ASK state for showing settings
```
solsarratea: Is is the expected state when showing one menu (scenario settings) or the other (creating-provider/line 261)?

src/planwise/component/scenarios.clj
```diff
@@ -62,7 +71,25 @@
...
+      (finally
+        (when raster
+          (let [path              (str "data/scenarios/" project-id "/")
+                scenario-filename (last (str/split raster #"/"))
+                condition-fn      (fn [file]
+                                    (when (and (.exists file) (not (.isDirectory file)))
+                                      (let [file-name (last (str/split (.getPath file) #"/"))
+                                            id        (re-find #"\d+" file-name)]
+                                        (= id (str (:id scenario))))))]
+            (delete-files-on-condition path condition-fn)))))))
```
ggiraldez: You should use the `scenario-filename` variable to build the names of the files to delete.

ggiraldez:
Ideally, you should also delete the coverage files from `coverage-cache` which correspond to providers in this scenario.

PR #498, opened 2018-09-27
Delete provider set

solsarratea:
Add feature for deleting provider set. 

>  "Delete provider-set given a provider-set-id.
     Providers and providers coverage referenced from provider-set are also deleted.
     When provider set is referenced from valid project an exception is thrown.
     If provider set is deleted all created files are deleted."

Changes:
UI/List of providers are updated when creating and deleting either a provider set, or a project.
Card in common/ui was adapted to show:
 -number of providers in provider set
 -number of projects depending on provider set
 -delete button when no projects depend on provider set.


src/planwise/component/providers_set.clj
```diff
@@ -244,6 +245,36 @@
...
+      (let [tx-store        (assoc-in store [:db :spec] tx)
+            provider-set-id {:provider-set-id provider-set-id}]
+        (delete-referenced-providers-coverage tx-store provider-set-id)
+        (delete-referenced-providers tx-store provider-set-id)
+        (delete-referenced-provider-set tx-store provider-set-id)))
+    (catch Exception e
+      (throw (ex-info "Provider set can not be deleted"
+                      {:provider-set-id provider-set-id}
+                      e)))
+    (finally (files/delete-files-recursively
```
ggiraldez: This should only be executed after a successful delete in the database.

PR #506, opened 2018-10-05
Fix numerical inputs and fix editable action

### LINKED ISSUE #503: Support for decimals numeric input is broken
Target population and capacity factor in project settings should support decimal input, but it doesn't work.

solsarratea:
Closes #503
Create general "mdc-input-field" which receives as input: props and component props. 
Props are attributes of the input element inside "mdc-input-field".
Component props are extra attributes for controlling local state of "mdc-input-field".

Besides, when editing an action:
 investment may now be float and suggested investment should not exceed remaining budget.

### DIFF
--- a/src/planwise/client/scenarios/edit.cljs
+++ b/src/planwise/client/scenarios/edit.cljs
@@ -50,6 +50,14 @@
     "increase-provider" (get-investment-from-project-config capacity increasing-costs)
     "create-provider"   (get-investment-from-project-config capacity building-costs)))
 
+(defn configured-costs?
+  [props]
+  (case (get-in props [:change :action])
+    "increase-provider" (not (empty? (:increasing-costs props)))
+    "upgrade-provider"  (and (pos? (:upgrade-budget props))
+                             (not (empty? (:increasing-costs props))))
+    (some? (:building-costs props))))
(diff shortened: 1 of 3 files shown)


src/planwise/client/scenarios/edit.cljs
```diff
@@ -65,10 +73,9 @@
         [common2/text-field {:label "Original capacity"
                              :read-only true
                              :value initial-capacity}])
-      [common2/numeric-text-field {:type "number"
-                                   :label (if increase? "Extra capacity" "Capacity")
-                                   :on-change  #(dispatch [:scenarios/save-key  [:changeset-dialog :change :capacity] %])
-                                   :value (or (:capacity change) "")}]
+      [common2/numeric-field {:label (if increase? "Extra capacity" "Capacity")
+                              :on-change  #(dispatch [:scenarios/save-key  [:changeset-dialog :change :capacity] %])
+                              :value (or (:capacity change) "")}]
```
ggiraldez: I don't think you need the `or` , since you're converting the value inside the `numeric-field` component. Besides, it's a bit weird for the value of a numeric field to be a string, but that's semantic nitpicking :)

PR #512, opened 2018-10-09
Prevent creating new providers coverage raster for points source demand

### LINKED ISSUE #511: Avoid creating raster coverages for point source scenarios
`engine/coverage-for-new-provider` will always write the coverage raster file, but for point source scenarios this is not necessary since covered sources are computed from the vectorized geometry.

solsarratea:
Checks whether project demand type is raster or not before creating raster-path for computing coverage
Closes #511

### DIFF
--- a/src/planwise/component/engine.clj
+++ b/src/planwise/component/engine.clj
@@ -83,7 +83,8 @@
    (coverage-for-new-provider coverage project change {}))
   ([coverage project {:keys [id location] :as change} geom-cache]
    (let [project-id  (:id project)
-         raster-path (new-provider-coverage-raster-path project-id id)
+         raster-path (when (= (:type-of-source-demand project) "raster")
+                       (new-provider-coverage-raster-path project-id id))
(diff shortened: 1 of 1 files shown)


src/planwise/component/engine.clj
```diff
@@ -83,17 +83,18 @@
    (coverage-for-new-provider coverage project change {}))
   ([coverage project {:keys [id location] :as change} geom-cache]
    (let [project-id  (:id project)
-         raster-path (new-provider-coverage-raster-path project-id id)
+         raster-path (when (= (:type-of-source-demand project) "raster")
+                       (new-provider-coverage-raster-path project-id id))
          cached-geom (get geom-cache id)]
      (if (and (some? cached-geom) (.exists (io/as-file raster-path)))
```
ggiraldez: This breaks if `raster-path` is `nil`, which will happen for source sets of type point. In this context, a `nil` raster path should be considered as a valid.

Issue #537, opened 2019-04-29
Split project creation/edition form in a multi step wizard


mverzilli:
For now stick to the exact same fields and controls in the current UI. We'll refine them iteratively.

PR #539, opened 2019-05-15
[WIP] Wizard

hdf1986:
Solves https://github.com/instedd/planwise/issues/537

This is a partial implementation but functional, we are currently working on a new iteration for this.

For now it should be ok to merge this code to staging (as far as the code seems ok...)

### DIFF
--- a/src/planwise/client/projects2/components/settings.cljs
+++ b/src/planwise/client/projects2/components/settings.cljs
@@ -55,6 +55,16 @@
              :on-click   (utils/prevent-default #(dispatch [:projects2/start-project (:id project)]))}
    (if (= (keyword (:state project)) :started) "Started ..." "Start")])
 
+
+(defn- project-next-step-button
+ [project step]
+ [m/Button {:id         "start-project"
+            :type       "button"
+            :unelevated "unelevated"
+            ; :disabled   (not (s/valid? :planwise.model.project/starting project))
+            :on-click   (utils/prevent-default #(dispatch [:projects2/next-step-project (:id project) step]))}
+            "Continue"])
+
 (defn- project-delete-button
   [state]
   [m/Button {:type     "button"
@@ -140,8 +150,8 @@
 (defn- current-project-step-consumers
   [read-only current-project]
   [:section {:class-name "project-settings-section"}
-   [section-header 2 "Demand"]
-   [sources-dropdown-component {:label     "Sources"
+   [section-header 2 "Consumers"]
+   [sources-dropdown-component {:label     "Consumer Dataset"
                                 :value     (:source-set-id current-project)
                                 :on-change #(dispatch [:projects2/save-key :source-set-id %])
                                 :disabled?  read-only}]
@@ -219,7 +229,7 @@
             "providers" [current-project-step-providers read-only @current-project @tags]
             "coverage" [current-project-step-coverage read-only @current-project]
             "actions" [current-project-step-actions read-only @current-project @build-actions @upgrade-actions]
-            "review" [:div "es review"]
+            "review" [:div "review"]
             [])]]])))
 
 (defn edit-current-project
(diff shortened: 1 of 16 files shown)


src/planwise/client/projects2/components/settings.cljs
```diff
@@ -125,92 +145,134 @@
...
         tags            (subscribe [:projects2/tags])]
-    (fn [{:keys [read-only]}]
-      [m/Grid {}
+    (fn [{:keys [read-only step]}]
+      [m/Grid {:class-name "wizard"}
+       [m/GridCell {:span 12 :class-name "steps"}
+        (doall
+         (map-indexed (fn [i iteration-step]
+                        [:a {:key i
+                             :class-name (join " " [(if (= iteration-step step) "active") (if (s/valid? (keyword (str "planwise.model.project-" iteration-step) "validation") @current-project) "complete")])
```
ggiraldez: Dynamically building the spec name at runtime is discouraged. A better approach would be to have a map from step to spec. Maybe even you can put more data in that map, like section title, link to next section, etc.

ggiraldez: On second read, you might want to have a vector which includes all this info. Eg.

```
(let [sections [{:step :goal :spec :planwise.model.project/goal :title "Goal"}
                {:step :consumers :spec :planwise.model.project/consumers :title "Consumers"}
                ...]]
  (map-indexed
    (fn [i section]
      [:a {:key section :class-name "..."} [:div (:title section)]]))
```

src/planwise/client/projects2/components/settings.cljs
```diff
@@ -125,92 +145,134 @@
...
-                            :list        @build-actions}]
-
-          [:div [:p [m/Icon "arrow_upward"] "Upgrading a provider so that it can satisfy demand would cost..."]]
-          [current-project-input "" [:config :actions :upgrade-budget] "number" {:disabled read-only :class "project-setting"}]
-
-          [:div [:p [m/Icon "add"] "Increase the capactiy of a provider by..."]]
-          [listing-actions {:read-only?   read-only
-                            :action-name :upgrade
-                            :list        @upgrade-actions}]]]]])))
+         (case step
```
ggiraldez: You can put the section component in the data structure describing the sections above as well and you won't need this case.

src/planwise/client/projects2/handlers.cljs
```diff
@@ -31,6 +39,23 @@
                      (assoc :list new-list))
       :navigate  (routes/projects2-show {:id project-id})})))
 
+(rf/reg-event-fx
+ :projects2/infer-step
+ in-projects2
+ (fn [{:keys [db]} [_ project]]
+   (let [steps ["goal", "consumers", "providers", "coverage", "actions", "review"]
```
ggiraldez: You can use the sections data structure here again and build a function which returns the first item for which the spec fails.

src/planwise/client/projects2/handlers.cljs
```diff
@@ -31,6 +39,23 @@
...
+         first-invalid-step (first (filter #(not (s/valid? (keyword (str "planwise.model.project-" %) "validation") project)) steps))
+         selected-step (if (not-empty first-invalid-step) first-invalid-step "review")]
+     {:navigate (routes/projects2-show-with-step {:id (:id project) :step selected-step})})))
+
+(rf/reg-event-fx
+ :projects2/next-step-project
+ in-projects2
+ (fn [{:keys [db]} [_ project-id step]]
+   (let [steps ["goal", "consumers", "providers", "coverage", "actions", "review"]]
+     {:navigate (routes/projects2-show-with-step {:id project-id :step (steps (inc (.indexOf steps step)))})})))
```
ggiraldez: `.indexOf` is from JS. This will not work in general for CLJS structures. You can have the next step as part of the section description.

ggiraldez: Btw, since this is the third time you have repeated the vector of steps, it's screaming for refactor :) Clearly the list of steps has more entity by itself.

src/planwise/model/project_review.cljc
```diff
@@ -0,0 +1,42 @@
...
+(s/def ::coverage (s/keys :req-un [::filter-options]))
+
+;; Actions
+(s/def ::budget number?)
+(s/def ::actions (s/keys :req-un [::budget]))
+
+;; Config
+(s/def ::config (s/keys :req-un [::demographics ::actions ::coverage ::providers]))
+
+(s/def ::validation (s/keys :req-un [::id ::owner-id ::name ::config ::provider-set-id ::source-set-id ::region-id]))
```
ggiraldez: You are duplicating a lot of code here. Define the leaf spec once and then make small sets for the different sections. Finally you can probably use something like https://clojuredocs.org/clojure.spec.alpha/merge to build the `review` spec.

PR #571, opened 2019-12-20
567 total demand under coverage

juanboca:


src/planwise/component/scenarios.clj
```diff
@@ -54,7 +54,15 @@
   (when (some? raster)
     (io/delete-file (io/file (str "data/" raster ".tif")) :silent)
     (io/delete-file (io/file (str "data/" raster ".map.tif")) :silent)
-    (io/delete-file (io/file (str "data/" raster ".coverage.tif")) :silent)))
+    (io/delete-file (io/file (str "data/" raster ".coverage.tif")) :silent)
+    (io/delete-file (io/file (str "data/" raster ".base-demand.tif")) :silent)
+
+    (let [old-provider-ids (set (keys (:new-providers-geom scenario)))
```
ggiraldez: This is obsolete code which was removed in c9f0c43c6d7861558e7b80946e6f985ec37f6ce6

client/src/planwise/client/scenarios/views.cljs
```diff
@@ -139,7 +140,7 @@
       "leaflet-circle-for-change"))})
 
 (defn simple-map
-  [{:keys [bbox]} scenario state error read-only?]
+  [{:keys [bbox]} scenario state error read-only? unit-name]
```
ggiraldez: This extra parameter doesn't feel right. It's only passed in to forward to `show-provider`, but make little sense. The first parameter is the project, and the demand unit can be obtained from it. But better would be to obtain it through a subscription. It doesn't make much sense to forward it as an isolated parameter to `show-provider` as well. I'd either merge it with the provider. Needless to say the code needed a huge refactor before this change...

ggiraldez: Also, for these types of reagent components, don't forget to add mirror the arguments list in the returned function below (line 152)

src/planwise/component/scenarios.clj
juanboca: My bad. This is a merge error

client/src/planwise/client/scenarios/views.cljs
juanboca: I don't know if it makes sense to have a subscription for something that will remain unchanged. I agree on extracting it on the simple-map function from the project instead of receiving it, but other solutions seem more complex and I don't see the benefit. 
Let's talk next week :)

PR #573, opened 2020-01-30
568 list of suggested facilities

juanboca:


client/src/planwise/client/scenarios/views.cljs
```diff
@@ -244,53 +257,88 @@
...
+      (if @suggested-locations
+        [:<>
+         [:div {:class-name "suggestion-list"}
+          [edit/create-new-action-component @view-state (or @computing-best-locations? @computing-best-improvements?)]]
+         [suggested-locations-list @suggested-locations state]]
+        [:<>
+         [:div
+          [scenario-info @view-state current-scenario unit-name]
+          [edit/create-new-action-component @view-state (or @computing-best-locations? @computing-best-improvements?)]
+          (if (or @computing-best-locations? @computing-best-improvements?)
```
ggiraldez: The expression `(or @computing-best-locations? @computing-best-improvements?)` is repeated 3 times. Extracting it to a `let` binding (eg. `computing-suggestions?`) would improve readability. You can also extract the action button component as well.

client/src/planwise/client/scenarios/views.cljs
```diff
@@ -244,53 +257,88 @@
...
+            [:div {:class-name "info-computing-best-location"}
+             [:small (if @computing-best-locations?
+                       "Computing best locations ..."
+                       "Computing best improvements...")]])]
+         [:div (when-not error {:class-name "fade"})]
+         (if error
+           [raise-alert current-scenario error]
+           [changeset/listing-component @providers-from-changeset])
+         [:div (when-not error {:class-name "fade inverted"})]]))))
```
ggiraldez: Both `div`s make sense only when `error` is falsey, so they can be moved inside the else position in the `if`.

PR #574, opened 2020-01-30
572 improve existing bug

juanboca:


juanboca:
Cherrypicked from `master` since the conflicts were very strange and hard to solve. 
It was more straightforward to just cherrypick

Issue #586, opened 2021-02-03
When a project analysis_type is action, measure as a cost or effort the amount of actions of the scenario.


manumoreira:

mmuller:
Working properly in version c184237a

Issue #584, opened 2020-12-18
Change wording in templates selection screen


manumoreira:
Plan ~~diagonostic devices~~ & sample referrals
Plan diagnostic services & sample referrals

PR #600, opened 2021-02-23
Feature/586 rename column investment to effort

ismaelbej:
* Rename investment column to effort #586
* Fix "suggestions to improve existing providers" that was broken when analysis-type is action
* Fix typo mentioned at #584

### DIFF
--- a/client/src/planwise/client/scenarios/changeset.cljs
+++ b/client/src/planwise/client/scenarios/changeset.cljs
@@ -9,7 +9,8 @@
             [planwise.client.utils :as utils]
             [planwise.client.routes :as routes]
             [clojure.string :as str]
-            [planwise.client.ui.rmwc :as m]))
+            [planwise.client.ui.rmwc :as m]
+            [planwise.common :as common]))
 
 (def action-icons
   {"create-provider" "domain"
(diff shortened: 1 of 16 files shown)


client/src/planwise/client/scenarios/changeset.cljs
```diff
@@ -44,7 +46,8 @@
      [m/Icon {} (get action-icons "create-provider")]
      [:div {:class-name "icon-list-text"}
       [:p {:class-name "strong"} (if name name (str "Suggestion " ranked))]
-      [:p {:class-name "grey-text"}  (str "Required Capacity: " action-capacity " Coverage: " coverage)]]]]
+      [:p {:class-name "grey-text"} (str "Required Capacity: " action-capacity
+                                         (if (not (nil? coverage)) (str " Coverage: " coverage)))]]]]
```
bcardiff, line 50: In which cases the coverage might be nil?

bcardiff (approved):
It looks great! I just have one doubt.

client/src/planwise/client/scenarios/changeset.cljs
ismaelbej: I'll check why coverage was nil.

ismaelbej: It is nil when requesting suggestions for an existing provider and it is not nil when requesting suggestions for a new provider. I checked in previous master and behavior is the same.

bcardiff: Ok, drop that as a comment in the code and merge. It might come back to haunt us that.

PR #601, opened 2021-02-26
Feature/588 sortable columns in scenarios list

ismaelbej:
Make scenarios list sortable.

### DIFF
--- a/client/resources/sass/site2.scss
+++ b/client/resources/sass/site2.scss
@@ -463,7 +463,7 @@ form.vertical, .fields-vertical {
         text-align: left;
         padding: 24px 20px;
         font-weight: 500;
-        font-size: 16px;
+        font-size: 1rem;
       }
     }
 
--- a/client/resources/sass/site2.scss
+++ b/client/resources/sass/site2.scss
@@ -5,6 +5,7 @@ $small-font-size: 0.81em;
 
 @import "material-components-web/material-components-web";
 @import "@material/theme/_color-palette";
+@import "../css/data-table";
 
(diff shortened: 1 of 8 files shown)


client/resources/sass/site2.scss
```diff
@@ -462,6 +463,7 @@ form.vertical, .fields-vertical {
         text-align: left;
         padding: 24px 20px;
         font-weight: 500;
+        font-size: 16px;
```
bcardiff, line 466:
```suggestion
        font-size: 1rem;
```

most of the styles are in rem units. relative to the root element font size.

client/src/planwise/client/design/views.cljs
```diff
@@ -64,6 +65,55 @@
       [:div.form-actions
        [m/Button {} "Continue"]]]]]])
 
+(defn scenarios-list-header
+  []
+  [:thead.rmwc-data-table__head
+   [:tr.rmwc-data-table__row.mdc-data-table__header-row
+    [:th]
+    [ui/sortable-table-header {:sortable true
```
bcardiff, line 73: Do we need a `:sortable true` if it's a `sortable-table-header` already?

client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -236,6 +236,21 @@
  (fn [db [_]]
    (update db :list asdf/invalidate!)))
 
+(rf/reg-event-db
+ :scenarios/change-sort-column
```
bcardiff, line 240: Wouldn't be simpler to have 1 event and letting the "change sort order on the current column" be another change sort column with column + order data?

bcardiff (approved):

PR #609, opened 2021-03-17
Feature/598 ux improve suggestions

ismaelbej:


client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -368,10 +373,11 @@
 (rf/reg-event-db
  :scenarios/edit-change
  in-scenarios
- (fn [db [_ change]]
+ (fn [db [_ change cancel-next-state]]
```
ggiraldez, line 376: The whole state management needs a full rewrite after careful thinking, but...

In this case you can omit this extra parameter as it should copy the current state, which you can read as `(:view-state db)`. You can use that to save the "previous" state that you want to restore after closing the dialog.

client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -133,10 +133,12 @@
  :scenarios/cancel-dialog
  in-scenarios
  (fn [db [_]]
-   (assoc db
-          :view-state :current-scenario
-          :changeset-dialog nil
-          :rename-dialog nil)))
+   (let [cancel-next-state (get-in db [:cancel-next-state])]
+     (assoc db
+            :view-state        (if (some? cancel-next-state) cancel-next-state :current-scenario)
```
ggiraldez, line 138: You can use `(or cancel-next-state :current-scenario)` here which should be equivalent in this case.

client/src/planwise/client/scenarios/views.cljs
```diff
@@ -192,8 +191,9 @@
             suggestions-layer       [:marker-layer {:points suggested-locations
                                                     :lat-fn #(get-in % [:location :lat])
                                                     :lon-fn #(get-in % [:location :lon])
-                                                    :popup-fn   #(show-suggested-provider % state)
+                                                    :popup-fn #(show-suggested-provider % state)
                                                     :icon-fn #(suggestion-icon-fn % selected-suggestion)
+                                                    :open-fn #(= % selected-suggestion)
```
ggiraldez, line 196: I was thinking something along the lines of adding an `:open?` attribute to each marker, but this works too. 

Ie.
```
:points (map #(assoc % :open? (= % selected-suggestion))
             suggested-locations)
```
This requires changing the `leaflet.core` code as well of course.

ggiraldez (commented):
I kinda like the idea of treating the view state as a stack. You could even formalize it a bit further by using a couple of functions `push-view-state` and `pop-view-state` (and of course get the current state through a subscription) to handle updates to the view state. I'm not sure it's worth it though, since as I said earlier the whole thing screams a full refactor.

Another thing you may want to explore is delaying the suggestion action selection to an event handler. Ie. always dispatch something generic like `:scenarios/edit-for-suggestion` and at the handler decide what to do (ie. create a new provider or edit the change). That would be save you from having to pass around the `action-fn` when rendering the suggestion list, plus I think it makes the code a bit clearer.

Take all these comments with an appropriate grain of salt. I don't know what is the timeline for the project so you may not have enough time to spend on these improvements.

ismaelbej:
@ggiraldez Refactoring view state as a stack will be discussed after evaluating the time required. All other suggestion were applied.

PR #614, opened 2021-03-30
Feature/613 fix map wizard

### LINKED ISSUE #613: Fix maps rendering in Wizard, the actual map layer is not showing up
Probably related to #612

ismaelbej:
Fixes #613

### DIFF
--- a/client/src/planwise/client/mapping.cljs
+++ b/client/src/planwise/client/mapping.cljs
@@ -3,9 +3,9 @@
             [re-frame.utils :as c]
             [planwise.client.config :as config]))
 
-(def emerald-mapbox-mapid "juanedi.177h17ed")
-(def bright-mapbox-mapid "juanedi/cis9iabkx002c31lg5awqyqa8")
-(def mapbox-access-token "pk.eyJ1IjoianVhbmVkaSIsImEiOiJFeVIxckN3In0.502Q6lu_hD-Bu3r9a0jUyw")
+(def emerald-mapbox-mapid "instedd/ckml2dykm4cl917o08mxux07b")
+(def bright-mapbox-mapid "instedd/ckml2dykm4cl917o08mxux07b")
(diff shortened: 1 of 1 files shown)


client/src/planwise/client/mapping.cljs
```diff
@@ -3,9 +3,8 @@
             [re-frame.utils :as c]
             [planwise.client.config :as config]))
 
-(def emerald-mapbox-mapid "juanedi.177h17ed")
-(def bright-mapbox-mapid "juanedi/cis9iabkx002c31lg5awqyqa8")
-(def mapbox-access-token "pk.eyJ1IjoianVhbmVkaSIsImEiOiJFeVIxckN3In0.502Q6lu_hD-Bu3r9a0jUyw")
+(def mapbox-mapid "instedd/ckml2dykm4cl917o08mxux07b")
+(def mapbox-access-token "pk.eyJ1IjoiaW5zdGVkZCIsImEiOiJja21ndHVrZ3cwMHQ5Mm9rZDgwaThkd3JoIn0.Kr3h9hO93IimCigXfYrBmw")
```
ggiraldez, line 7: Small suggestion, maybe not worth the effort: it'd be ideal if these values came from the environment configuration. The mechanism we have right now is:
1. define the variables in the application configuration with values from the environment, as in https://github.com/instedd/planwise/blob/master/resources/planwise/config.edn#L157 or https://github.com/instedd/planwise/blob/master/resources/planwise/config.edn#L99
2. add the values to the rendered client configuration in the home endpoint, by extracting the values from the Duct configuration map (eg. https://github.com/instedd/planwise/blob/master/src/planwise/endpoint/home.clj#L34). This will result in the values being rendered in the global JS variables `_CONFIG`
3. implement a couple of vars in the client's config namespace: https://github.com/instedd/planwise/blob/master/client/src/planwise/client/config.cljs#L14
4. use the vars from such namespace, as in https://github.com/instedd/planwise/blob/master/client/src/planwise/client/design/views.cljs#L139

This way we can define the map ID and access tokens as environment variables in the deployments and if they change we don't need to rebuild the Docker image.

Issue #621, opened 2021-04-07
Refactor view state


manumoreira:

ggiraldez:
This was done in this last round of development. 
See https://github.com/instedd/planwise/blob/518542978c1fe5bcf0de845dd8006ddf5c0c6aca/client/src/planwise/client/scenarios/db.cljs#L5 Dialogs states are also now independent of the UI state.

PR #622, opened 2021-04-07
Refactor view-state into a stack

ismaelbej:


client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -32,6 +32,40 @@
...
+ (fn [db [_]]
+   (println [".......z" (:views-stack db)])
+   (update-in db [:views-stack] #(if (empty? %) % (pop %)))))
+
+(rf/reg-event-db
+ :scenarios/reset-state
+ in-scenarios
+ (fn [db [_]]
+   (assoc-in db [:views-stack] [])))
+
```
ggiraldez, line 56: Manipulating the stack does not need to be done in separate events. You can extract the push/pop/reset functions, make them take the `db` as a first parameter and thread them with the rest of the `db` updates in the existing events. Eg.
```
(defn push-view
  [db new]
  (update-in db [:views-stack] #(conj % view)))

(defn pop-view
  [db]
  (update-in db [:views-stack] #(if (empty? %) % (pop %))))

(defn reset-view-state
  [db]
  (assoc-in db [:views-stack] []))
```

Then use it, for example:
```
(rf/reg-event-fx
  :scenarios/cancel-dialog
  (fn [{:keys [db]} [_]]
   {:db (-> db
            (assoc :changeset-dialog nil
                   :rename-dialog nil)
            (pop-view))}))
```
In short, make your own view stack manipulation functions and use them to mutate the db.

client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -32,6 +32,40 @@
...
+  [{:keys [views-stack]}]
+  (if (empty? views-stack)
+    :current-scenario
+    (peek views-stack)))
+
+(defn peek-view-1
+  [{:keys [views-stack]}]
+  (if (< (count views-stack) 2)
+    :current-scenario
+    (peek (pop views-stack))))
```
ggiraldez, line 67: This function is not used, correct?

client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -95,11 +129,10 @@
...
  in-scenarios
  (fn [{:keys [db]} [_ id]]
-   {:db  (assoc db :view-state :current-scenario)
-                                        ; reset view state to close any pending dialogs
-    :api (assoc (api/load-scenario id)
+   {:api (assoc (api/load-scenario id)
                 :on-success [:scenarios/save-current-scenario-and-track]
-                :on-failure [:scenarios/scenario-not-found])}))
+                :on-failure [:scenarios/scenario-not-found])
+    :dispatch [:scenarios/reset-state]}))
```
ggiraldez, line 135: Applying my suggestion, this would be:

```
{:api (assoc (api/load-scenario id)
                :on-success [:scenarios/save-current-scenario-and-track]
                :on-failure [:scenarios/scenario-not-found])}))
                :on-failure [:scenarios/scenario-not-found])
 :db (reset-view-state db)}
```

client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -110,35 +143,32 @@
...
+(rf/reg-event-fx
  :scenarios/clear-current-scenario
  in-scenarios
- (fn [db [_]]
-   (assoc db
-          :current-scenario nil
-          :view-state :current-scenario)))
+ (fn [{:keys [db]} [_]]
+   {:db       (assoc db :current-scenario nil)
+    :dispatch [:scenarios/reset-state]}))
```
ggiraldez, line 151: And this would be:
```
{:db (-> db
      (assoc :current-scenario nil)
      (reset-view-state))}
```

client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -110,35 +143,32 @@
...
-   (let [cancel-next-state (get-in db [:cancel-next-state])]
-     (assoc db
-            :view-state        (or cancel-next-state :current-scenario)
-            :cancel-next-state nil
-            :changeset-dialog  nil
-            :rename-dialog     nil))))
+ (fn [{:keys [db]} [_]]
+   {:db (assoc db
+               :changeset-dialog nil
+               :rename-dialog nil)
```
ggiraldez, line 170: These two look like they belong with the view state. I think it would be safe to move the value resets to the `reset-view-state` function.

client/src/planwise/client/scenarios/handlers.cljs
```diff
@@ -296,89 +322,95 @@
  :scenarios.new-action/toggle-options
```
ggiraldez, line 322: This handler is doing too much. At the very least I'd try to separate the logic for cancelling the ongoing computations from opening/closing the options menu. Of course you'll need to change the view code as well, but that should be straightforward. Having done that, this should be trivial to implement: if the view state is `show-options-to-create-provider`, then pop the view stack. Otherwise, push the `show-options-to-create-provider` to the stack. For cancelling the computations there already is an existing handler `abort-fetching-suggestions`  but that requires an extra argument. You can either dispatch that event from the view directly (with the appropriate argument), or change the handler to find which of the computations is active.

client/src/planwise/client/scenarios/edit.cljs
```diff
@@ -163,7 +163,7 @@
    [m/Fab {:class-name "btn-floating"
            :on-click #(dispatch [:scenarios.new-action/toggle-options])}
     (cond computing? "stop"
-          (or (= state :new-provider) (= state :new-intervention)) "cancel"
+          (or (= state :new-provider) (= state :new-intervention) (= state :new-suggestion)) "cancel"
```
ggiraldez, line 166: There are a couple of places where this or a similar pattern is repeated. I suggest extracting the predicate to a function (eg. `state-cancellable?` maybe).

client/src/planwise/client/scenarios/handlers.cljs
ismaelbej: I'll do it.

client/src/planwise/client/scenarios/handlers.cljs
ismaelbej: No, it isn't used anymore. I needed in some case but it didn't work.

ggiraldez:
The view state was refactored in this last round of development. We didn't model it as a stack in the end. See https://github.com/instedd/planwise/issues/621#issuecomment-992545149. Most of the work is in https://github.com/instedd/planwise/pull/692

PR #641, opened 2021-04-27
Feature/620 multiple transport coverage

ismaelbej:


cpp/walking-coverage.cpp
```diff
@@ -964,15 +994,30 @@ int main(int argc, char *argv[])
...
+    unique_ptr<float[]> new_cost =
+      run_dijkstra_on_friction_layer(friction.get(),
+                                     width, height, nodata,
+                                     frictionRaster.pixel_width_meters(),
+                                     frictionRaster.pixel_height_meters(),
+                                     pixelOrigin.first, pixelOrigin.second,
+                                     options._maxTimeCost[i],
+                                     options._minFriction[i]);
+
+    // Merging layers scale costs to make it is easier to calculate the isochrone
```
ggiraldez, line 1018: This comment is a bit misleading. I'd say you need to adjust the new cost layer to take it to the same scale as the original in order to compute the isochrone.

client/src/planwise/client/coverage.cljs
```diff
@@ -81,15 +81,20 @@
 
 (defn coverage-algorithm-filter-options
   [{:keys [coverage-algorithm value on-change empty disabled?]}]
-  (let [list      (rf/subscribe [:coverage/supported-algorithms])
-        criteria  (get-in @list [(keyword coverage-algorithm) :criteria])]
+  (let [list         (rf/subscribe [:coverage/supported-algorithms])
+        criteria     (get-in @list [(keyword coverage-algorithm) :criteria])
+        valid-keys   (keys criteria)
+        update-value (fn [key change]
+                       (if (= coverage-algorithm "drive-walk-friction")
```
ggiraldez, line 88: I don't think you need this `if` here (plus the special case handling should is a smell). In fact, the previous behavior was wrong in that it only allowed a single criterion. Filtering the keys (using `select-keys`) is the correct and general solution.

client/src/planwise/client/projects2/components/settings.cljs
```diff
@@ -232,20 +232,31 @@
      [:label "Tags: " [tag-set @tags read-only]]
      [count-providers @tags @current-project provider-unit]]))
 
+; Accept previous coverage algorithms as the new one combining both
+(defn- get-coverage-algorithm
+  [coverage]
+  (case coverage
+    "driving-friction" "drive-walk-friction"
+    "walking-friction" "drive-walk-friction"
+    coverage))
```
ggiraldez, line 241: I think we should do this migration server side given that the list of supported algorithms now only returns `drive-walk-friction`, eg. in `planwise.component.projects2/get-project`. You _could_ probably remove the old cases of `compute-coverage-polygon` in that case as well.

ggiraldez (commented):
This looks great! But, please do check the comments, especially regarding the keys to the default `frictions` var.

src/planwise/component/coverage.clj
```diff
@@ -79,21 +80,48 @@
   (let [db-spec         (:spec db)
         friction-raster (friction/find-friction-raster db-spec coords)
         max-time        (:walking-time criteria)
-        min-friction    (float (/ 1 100))]
+        min-friction    (:walking-friction frictions)]
```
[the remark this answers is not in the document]
ismaelbej: I'll fix it.

src/planwise/component/coverage.clj
```diff
@@ -79,21 +80,48 @@
...
+                                 :friction [min-friction]})
       (throw (ex-info "Cannot find a friction raster for the given coordinates" {:coords coords})))))
 
 (defmethod compute-coverage-polygon :driving-friction
   [{:keys [db runner]} coords criteria]
   (let [db-spec         (:spec db)
         friction-raster (friction/find-friction-raster db-spec coords)
         max-time        (:driving-time criteria)
-        min-friction    (float (/ 1 2000))]
+        min-friction    (:driving-friction frictions)]
```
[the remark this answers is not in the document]
ismaelbej: I'll fix it.

cpp/walking-coverage.cpp
ismaelbej: I'll write a better explanation.

client/src/planwise/client/coverage.cljs
ismaelbej: I'll fix it.

client/src/planwise/client/projects2/components/settings.cljs
ismaelbej: I'll do it.

PR #642, opened 2021-04-28
Feature/634 filter consumers source

ismaelbej:


client/src/planwise/client/projects2/handlers.cljs
```diff
@@ -11,6 +11,12 @@
 
 (def in-projects2 (rf/path [:projects2]))
 
+(defn- reset-state
```
ggiraldez, line 14: This name is too generic. What do you think about `clear-current-project`?

client/src/planwise/client/projects2/components/settings.cljs
```diff
@@ -183,17 +183,39 @@
                                   :on-change #(dispatch [:projects2/save-key :region-id %])
                                   :model     (:region-id @current-project)
                                   :disabled? read-only}]]))
+
+(defn- set-source-types
```
ggiraldez, line 187: I'd call this `toggle-source-type` instead, and you shouldn't need to check the DOM node's `checked` attribute. You can check the current value of `sources` instead.

client/src/planwise/client/projects2/handlers.cljs
```diff
@@ -97,6 +103,7 @@
  in-projects2
  (fn [db [_ current-project]]
    (-> db
+       (reset-state)
```
ggiraldez, line 106: This looks a bit out of place here. Is it intended?

ggiraldez (approved):
Looks good, but check the comments for some suggestions.

client/src/planwise/client/projects2/components/settings.cljs
ismaelbej: I'll do it.

client/src/planwise/client/projects2/handlers.cljs
ismaelbej: It seems more appropiate.

client/src/planwise/client/projects2/handlers.cljs
ismaelbej: The intention was to clear the source-types selection when switching projects. Perhaps it is a bit nicer to select the correct type from the source.

client/src/planwise/client/sources/components/dropdown.cljs
```diff
@@ -10,9 +10,13 @@
 
 (rf/reg-sub
```
ismaelbej, line 11: @ggiraldez I'd like to have a second opinion about the sources dropdown. I moved the filter by type from`sources-dropdown-component` to the subscription `:sources/dropdown-options`, while doing that I used `:sources/list` instead of directly reading from the `db`, removed calls to `asdf/should-reload?`.

PR #646, opened 2021-05-03
Feature/633 project settings

ismaelbej:


client/src/planwise/client/projects2/components/settings.cljs
```diff
@@ -382,7 +422,8 @@
           (map-indexed (fn [i iteration-step]
                          [:a {:key i
                               :class-name (join " " [(if (= (:step iteration-step) step) "active") (if (s/valid? (:spec iteration-step) project) "complete")])
-                              :href (routes/projects2-show-with-step {:id (:id project) :step (:step iteration-step)})}
+                              :href ((if read-only routes/projects2-settings-with-step routes/projects2-show-with-step)
+                                     {:id (:id project) :step (:step iteration-step)})}
                           (if (s/valid? (:spec iteration-step) project) [m/Icon "done"] [:i (inc i)])
                           [:div (:title iteration-step)]]) sections)]
          [m/GridCell {:span 6}
```
ggiraldez, line 429: There is a `(dispatch [:projects2/infer-step @current-project])` below here that is out of place and it is what was causing the navigation back to the scenarios list before this patch. You fixed this behavior by providing a default `review` step when showing the settings. But the step inference is a pure computation. So we can improve and simplify the code a bit further by removing the `dispatch` below, and inferring the current step if `nil` (using the `infer-step` function if editing or defaulting to `review` if showing) before looking for the `step-data`.

ggiraldez (approved):
The changes look good, but I think further improvement and simplifications are possible. Step inference should not be side-effecting during project setup. I'm finding it difficult to explain myself, so please let's talk if it's not clear what I mean.

client/src/planwise/client/projects2/components/settings.cljs
ismaelbej: @ggiraldez I removed the infer-step but I can't get rid of the dispatch. If the URL doesn't change then a step may become valid while typing and it will suddenly jump to the next step.

client/src/planwise/client/projects2/components/settings.cljs
```diff
@@ -392,13 +393,28 @@
...
+  {"goal"      current-project-step-goal
+   "consumers" current-project-step-consumers
+   "providers" current-project-step-providers
+   "coverage"  current-project-step-coverage
+   "actions"   current-project-step-actions
+   "review"    current-project-step-review})
+
+(defn step->spec
+  [step]
+  (keyword "planwise.model.project" (str step "-step")))
```
ggiraldez, line 406: This is fine, but I would have left the specs a part of the `core/sections` definition. That way the spec is not necessarily fixed from the step name.

client/src/planwise/client/projects2/components/settings.cljs
```diff
@@ -392,13 +393,28 @@
...
-   {:step "review" :title "Review" :component current-project-step-review :spec :planwise.model.project/review-step :next-step nil}])
+  (->> core2/sections
+       (#(partition (count %) 1 (repeat nil) %))
+       (apply map (fn [x y] [x y]))
+       (map (fn [[step next]]
+              (assoc step
+                     :component (get step->component (:step step))
+                     :spec (step->spec (:step step))
+                     :next-step (:step next))))
+       (into [])))
```
ggiraldez, line 417: `next-step` can also be part of the `core2/sections` definition. But, another way of doing this:
```clojure
(mapv (fn [step next]
         (assoc step ...))
      core2/sections  ; the original list
      (concat (drop 1 core2/sections) (repeat nil))) ; the list "shifted" by 1 element
```

client/src/planwise/client/projects2/components/settings.cljs
```diff
@@ -410,12 +426,19 @@
...
             region-geo      (subscribe [:regions/preview-geojson (:region-id @current-project)])
             preview-map-url (if (:region-id @current-project) (static-image @region-geo map-preview-size))]
         (dispatch [:regions/load-regions-with-preview [(:region-id @current-project)]])
+        (when (nil? step-data)
+          (dispatch
+           [(if read-only :projects2/navigate-to-settings-project :projects2/navigate-to-step-project)
+            (:id project)
+            (or (when-not read-only
+                  (:step (first (filter #(not (s/valid? (:spec %) project)) sections))))
+                (:step (last sections)))]))
```
ggiraldez, line 441: This can be extracted to a function `infer-step` and that can make the code easier to understand.

PR #661, opened 2021-05-14
Feature/659 Scenario in invalid state

ismaelbej:
Disable the "create new scenario" button and "new provider" menu if a scenario is in a pending state.


ggiraldez (approved):
Looks great!

PR #672, opened 2021-11-15
Add coverage % to each action in the table view for scenarios

### LINKED ISSUE #668: Show progress in coverage in action table
Add a column showing the % of coverage for each action
screencapture-planwise-instedd-org-projects2-232-scenarios-568-2021-11-05-11_21_36

ggiraldez:
Closes #668

The percentage shown for each action is the new demand covered by the given action over the total demand in the scenario. The sum of the coverage percentages for all actions should equal the percentage increase for the whole scenario.

Minor tweaks to styles:
- move settings button to the right of the scenario title
- reduce margins a bit
- change the cursor on the sidebar expand button to a pointer

Screenshot from 2021-11-15 17-04-35
Screenshot from 2021-11-15 17-04-48

### DIFF
--- a/client/src/planwise/client/scenarios/changeset.cljs
+++ b/client/src/planwise/client/scenarios/changeset.cljs
@@ -81,8 +81,7 @@
         suggestions)])
 
 (defn- changeset-table-row
-  [{:keys [source-demand] :as props}
-   {:keys [name change satisfied-demand reachable-demand] :as provider}]
+  [{:keys [source-demand] :as props} {:keys [name satisfied-demand] :as provider}]
   [:tr
    [:td.col-action-icon [provider-icon provider]]
    [:td.col-action-name name]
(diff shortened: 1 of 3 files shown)


client/src/planwise/client/scenarios/changeset.cljs
```diff
@@ -81,11 +81,13 @@
         suggestions)])
 
 (defn- changeset-table-row
-  [props {:keys [name change] :as provider}]
+  [{:keys [source-demand] :as props}
+   {:keys [name change satisfied-demand reachable-demand] :as provider}]
```
ismaelbej, line 85: reachable-demand is used so it could be removed here

ismaelbej (approved):
Looks good to me!

client/src/planwise/client/scenarios/changeset.cljs
ggiraldez: Removed `reachable-demand` and `change` as well since they were not used. Thanks!

PR #719, opened 2024-03-06
CI on Github Actions

matiasgarciaisaia:
This PR moves the CI from CircleCI to Github Actions.

The only breaking change we introduce is that the `instedd/planwise-mapserver` images are now tagged `VERSION-mapserver` and `VERSION-mapcache` instead of `mapcache-VERSION` like before.


docker-compose.ci.yml
```diff
@@ -2,9 +2,6 @@ version: '2'
 
 services:
   app:
-    volumes:
-      - .:/app
-      - ~/cache/.m2:/root/.m2
```
ggiraldez, line 7: The Maven cache would be good to restore at some point. The overhead of downloading the same dependencies over and over is high.

mapserver/deploy.sh
```diff
@@ -1,22 +0,0 @@
...
-
-if [ $# -lt 1 ]; then
-  echo "Usage: $0 DOCKER_TAG"
-	exit 1
-fi
-
-TAG=${1/\//_}
-
-MAPCACHE_TAG=instedd/planwise-mapserver:mapcache-$TAG
-MAPSERVER_TAG=instedd/planwise-mapserver:mapserver-$TAG
```
ggiraldez, line 16: Oh, I see why you added the tag suffix above. I think in this case the `TAG` was not meant to be a versioning tag, but a content tag. Ie. at some point we would generate the mapserver image with some regional TIFFs included. But I don't think we've used them in that way for a very long time.

ggiraldez (approved):
Looks good to me! I left a couple of minor comments.

docker-compose.ci.yml
matiasgarciaisaia: I think we should _eventually_ split the image in two - one with the base dependencies, and the other one with that PLUS the app's code.

mapserver/deploy.sh
matiasgarciaisaia: Interesting.

I see [in Docker Hub](https://hub.docker.com/r/instedd/planwise-mapserver/tags) we have some tag `kenya-mapserver-stable`, but I also see `mapcache-latest`, `mapcache-master` & `mapcache-stable` (created a few years ago).

I mean, I really am not sure about what these things are or do, so I'll trust your judgment here. I'm just replicating the CI we had.

ggiraldez: Oh, wait. I didn't realize that we didn't have a different repo for mapcache. That's what the tag is being used for. I'd go for a different Docker repo though.

matiasgarciaisaia: Agree on that. I'll leave that for later, though.

PR #722, opened 2024-03-11
Consumers autoselect

### LINKED ISSUE #717: Enhance consumers selection
### Use case
When created #715 we realized that the selection of consumers is a bit messy in the UI

consumersDrop_1

We should find a way to make the UI simpler for this task

Some options would be:

1. Order the dropdown by country name (so at lest the user decide with population source to use)
2. Filter the population sources based on the region selection in the wizard
3. Replace the dropdown with a search-box

matiasgarciaisaia:
Replace the `Consumers Dataset` dropdown with an autoselect control like the one we have for Region selection, with a filter.

Dataset selector showing the list of datasets matching the letter "a"

Fixes #717

### DIFF
--- a/client/src/planwise/client/sources/components/dropdown.cljs
+++ b/client/src/planwise/client/sources/components/dropdown.cljs
@@ -4,6 +4,7 @@
             [planwise.client.asdf :as asdf]
             [planwise.client.components.common2 :as common2]
             [planwise.client.utils :as utils]
+            [planwise.client.ui.filter-select :as filter-select]
             [planwise.client.ui.rmwc :as m]))
 
 (def in-sources (rf/path [:sources]))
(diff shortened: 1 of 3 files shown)


mverzilli (approved):
LGTM!

client/src/planwise/client/sources/components/dropdown.cljs
```diff
@@ -14,29 +15,17 @@
...
-                :on-change    on-change}]))
\ No newline at end of file
+  [attrs]
+  (let [props (merge {:choices   @(rf/subscribe [:sources/dropdown-options])
+                      :label-fn  :name
+                      :render-fn (fn [source] [:div.option-row
+                                               [:span (:name source)]
+                                               [:span.option-context (:value source)]])}
+                     attrs)]
+    (into [filter-select/single-dropdown] (mapcat identity props))))
```
matiasgarciaisaia, line 31: This view is essentially the same as the one for the `regions-dropdown-component`, so I guess there's a component we should extract - but I didn't trust my ClojureScript-foo enough to go for it now.

Issue #724, opened 2024-03-29
Providers autoselect


manumoreira:
Currently regions and sources have an autoselect in the UI, let's align the providers selection to that.
The providers selection (3rd step in the project wizard) should have the same autoselect control used for regions and sources.

PR #725, opened 2024-04-02
Use autoselect control for Providers

matiasgarciaisaia:
So it matches both the Region and Consumers pickers.

Screenshot of the component showing two fake-sample entries being filtered

See #724

### DIFF
--- a/client/src/planwise/client/providers_set/components/dropdown.cljs
+++ b/client/src/planwise/client/providers_set/components/dropdown.cljs
@@ -1,33 +1,12 @@
 (ns planwise.client.providers-set.components.dropdown
-  (:require [re-frame.core :refer [subscribe dispatch]]
-            [reagent.core :as r]
-            [clojure.string :as string]
-            [planwise.client.asdf :as asdf]
-            [planwise.client.utils :as utils]
-            [planwise.client.components.common2 :as common2]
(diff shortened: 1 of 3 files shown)


client/src/planwise/client/providers_set/components/dropdown.cljs
```diff
@@ -1,33 +1,16 @@
...
-                    common2/disabled-input-component
-                    providers-set-select-component)]
+  [attrs]
+  (let [list  (rf/subscribe [:providers-set/list])
+        props (merge {:choices   @(rf/subscribe [:providers-set/dropdown-options])
+                      :label-fn  :label
+                      :render-fn (fn [provider-set] [:div.option-row
+                                               [:span (:label provider-set)]])}
+                     attrs)]
     (when (asdf/should-reload? @list)
```
matiasgarciaisaia, line 14: I don't know in which scenarios we should reload the list, but it was needed to _at least_ load the options upon page load, so I kept this here.

manumoreira: It's pretty usual to realize there's a providers source missing when you get to that screen so, if I get this right, it is good to have a reload after page load

mverzilli (approved):

PR #732, opened 2024-05-31
Use mapshaper to simplify a region geometry

### LINKED ISSUE #727: Simplify region shapes in projects
After using Planwise in several projects it was pretty clear that the quality of the region shapes (polygons) are not aligned with the quality (level of detail) of the population and accessibility (friction) sources. In other words, we have super detailed region shapes combined with more coarse population and friction rasters.

We might consider using the `mapshaper -simplify` command to get a balanced quality of the 3 sources.
Here's an example of size for simplification https://github.com/instedd/planwise/blob/7a102523feca0eb584180a43bc26486205766d0c/scripts/tools/update-region-previews#L59

The main issue with having a highly detailed region shape is that it makes operating with it (eg. for filtering of providers, clipping the demand rasters) very slow while not providing any actual value. 

From the slack conversation

> The most complicated part would be adding a simplification step when importing the regions and choose some criteria for the precision cut.
We have a size criteria implemented [here](https://github.com/instedd/planwise/blob/7a102523feca0eb584180a43bc26486205766d0c/scripts/tools/update-region-previews#L59) which is used for the previews. I would prefer a resolution criteria for this use case, and I'm sure it should be easily doable. Just need to stitch it together.
It may be a good time to migrate the remaining Bash scripts to Babashka. Or at least those involved here: [load-regions](https://github.com/instedd/planwise/blob/7a102523feca0eb584180a43bc26486205766d0c/scripts/population/load-regions) and [load-friction-raster](https://github.com/instedd/planwise/blob/7a102523feca0eb584180a43bc26486205766d0c/scripts/friction/load-friction-raster).

ismaelbej:
* Modify load-regions to use mapshaper to simplify a region using the distance parameter
* Update load-friction-raster to pass the -cblend 1 option
* Add `-f` parameter to `load-regions` to update an existing region
* Set `INTERVAL_DISTANCE=100` as default distance for simplification
* Document new parameters in scripts/population/README.md

Closes #727

### DIFF
--- a/scripts/population/load-regions
+++ b/scripts/population/load-regions
@@ -48,7 +48,7 @@ fi
 
 DATA_PATH=${DATA_PATH:-/data}/geojson
 
-mkdir -p ${DATA_PATH}
+mkdir -p "${DATA_PATH}"
 
 # if the folder (still) doesn't exist then exit with error
 if [[ ! -e ${DATA_PATH}/$COUNTRY_CODE ]]; then
@@ -77,38 +77,18 @@ for i in $LEVELS; do
     # Simplify using distance
     OFILE=${DATA_PATH}/${COUNTRY_CODE}/${COUNTRY_CODE}_adm${i}_${INTERVAL_DISTANCE}.geojson
 
-    mapshaper $FILE -simplify interval=$INTERVAL_DISTANCE -o $OFILE
+    mapshaper "$FILE" -simplify interval=$INTERVAL_DISTANCE -o "$OFILE"
 
-    SIZE=$(stat --printf="%s" $FILE)
-    OSIZE=$(stat --printf="%s" $OFILE)
+    SIZE=$(stat --printf="%s" "$FILE")
+    OSIZE=$(stat --printf="%s" "$OFILE")
 
     echo "   -> Finished file size went from $SIZE to $OSIZE"
   fi
 
-  if [ "$FORCE" = "1" ]; then
-    psql -q -d $POSTGRES_DB -U $POSTGRES_USER -h $POSTGRES_HOST -p $POSTGRES_PORT << SQL_SCRIPT
+  psql -q -d $POSTGRES_DB -U $POSTGRES_USER -h $POSTGRES_HOST -p $POSTGRES_PORT << SQL_SCRIPT
 
-      WITH data AS (SELECT \$$`cat $OFILE`\$$::json AS fc)
-      UPDATE "regions" SET the_geom=f.the_geom
-      FROM (
-        SELECT
-          '${COUNTRY_NAME}' as country,
-          feat#>>'{properties,name}' AS name,
-          ${i} as admin_level,
-          ST_SetSRID(ST_CollectionExtract(ST_Multi(ST_GeomFromGeoJSON(feat->>'geometry')), 3), 4326) as the_geom
-        FROM (
-          SELECT json_array_elements(fc->'features') AS feat
-          FROM data
-        ) AS dummy
-      ) AS f
-      WHERE regions.country=f.country AND regions.name=f.name AND regions.admin_level=f.admin_level;
-
-SQL_SCRIPT
-  else
-    psql -q -d $POSTGRES_DB -U $POSTGRES_USER -h $POSTGRES_HOST -p $POSTGRES_PORT << SQL_SCRIPT
-
-      WITH data AS (SELECT \$$`cat $OFILE`\$$::json AS fc)
-      INSERT INTO "regions" (country, name, admin_level, the_geom)
+    WITH data AS (SELECT \$$`cat "$OFILE"`\$$::json AS fc)
+    INSERT INTO "regions" (country, name, admin_level, the_geom)
       SELECT
         '${COUNTRY_NAME}',
         feat#>>'{properties,name}' AS name,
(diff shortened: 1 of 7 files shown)


scripts/population/load-regions
```diff
@@ -33,13 +33,30 @@ if [[ ! -e ${DATA_PATH}/$COUNTRY_CODE ]]; then
   exit 1
 fi
 
+# Simplification parameter
+# From mapshaper wiki https://github.com/mbloch/mapshaper/wiki/Command-Reference#-simplify
+#
+# interval= Specify simplification amount in units of distance.
+#           Uses meters when simplifying unprojected datasets in 3D space (see planar option below),
+#           otherwise uses the same units as the source data.
+INTERVAL_DISTANCE=${INTERVAL_DISTANCE-1000}
```
ggiraldez, line 42: Could you please expand the comment above explaining that in our case since our GeoJSON are not projected (ie. use WGS84 coordinates) then this means the resolution for the simplified shape will be 1km? (Or at least that's what I understand from the documentation :grin:)

ggiraldez (commented):
This works but will not reprocess the existing regions (if I understand correctly) and we can't delete them and re-add them because they already have referencing projects.

It would also be nice to see the resulting size of the simplified shape, to adjust the interval if deemed necessary. For large countries (or large provinces) the default threshold may not be enough to reduce the size enough and still hog the CPU when performing spatial queries.

ismaelbej:
I'll add a `-f` to force updating the geometries.

scripts/population/load-regions
ismaelbej: I wasn't sure about what the parameter was measuring in this case. I compared the mapshaper output of the Kenya map or 100 and 1000 (it is the island on the south east)

Screenshot 2024-05-30 at 10 42 48 PM

Screenshot 2024-05-30 at 10 46 03 PM

ggiraldez: Hmm... I'm thinking this can potentially lead to legitimate providers being excluded. Maybe we should play a bit with the thresholds and the problematic countries and check the resulting sizes. We may get away with a smaller threshold. @manumoreira what was the country that triggered the original issue?

ggiraldez: In fact, a default of 100 may work fine. I'm testing locally with some datasets I have and Philipines shrinks from 58 Mb to 2.5 Mb using 100 as the threshold. That should be acceptable I think, since Kenya's contour is 4.9 Mb unprocessed and we didn't have problems with that country so far. NB. sizes are of the admin level 0 `.geojson` files.

manumoreira: The country was Mozambique

ismaelbej: Comparison between the original and the 100 meters simplification

Screenshot 2024-05-31 at 10 14 44 AM

ismaelbej: I had to zoom more because the difference wasn't visible from https://mapshaper.org/.

scripts/population/load-regions
```diff
@@ -33,25 +56,71 @@ if [[ ! -e ${DATA_PATH}/$COUNTRY_CODE ]]; then
...
-      SELECT json_array_elements(fc->'features') AS feat
-      FROM data
-    ) AS f;
+
+  OFILE=$FILE
+  if [[ ! -z $INTERVAL_DISTANCE ]]; then
+    # Simplify using distance
+    OFILE=${DATA_PATH}/${COUNTRY_CODE}/${COUNTRY_CODE}_adm${i}_${INTERVAL_DISTANCE}.geojson
+
+    mapshaper $FILE -simplify interval=$INTERVAL_DISTANCE -o $OFILE
```
matiasgarciaisaia, line 80: I think we kind of control the paths, so not really an issue, but it wouldn't hurt to escape some of this variables.

scripts/population/load-regions
```diff
@@ -33,25 +56,71 @@ if [[ ! -e ${DATA_PATH}/$COUNTRY_CODE ]]; then
...
+      INSERT INTO "regions" (country, name, admin_level, the_geom)
+      SELECT
+        '${COUNTRY_NAME}',
+        feat#>>'{properties,name}' AS name,
+        ${i},
+        ST_SetSRID(ST_CollectionExtract(ST_Multi(ST_GeomFromGeoJSON(feat->>'geometry')), 3), 4326) as the_geom
+      FROM (
+        SELECT json_array_elements(fc->'features') AS feat
+        FROM data
+      ) AS f;
```
matiasgarciaisaia, line 120: Could this be turned into a single "UPSERT" (ie, [`INSERT ... ON CONFLICT UPDATE`](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT))? Forcing usually means we want to reach the end state regardless of the actual state - ie, don't care if the data exists or not, we want it to exist afterwards. So forcing on a non existent region should also succeed.

matiasgarciaisaia (commented):
Can't comment further on the task being done - I'll trust Gus' opinion on that.

scripts/population/load-regions
ismaelbej: I understand that the upset will require an index on (country,name,admin_level). Is that OK?

scripts/population/load-regions
ismaelbej: Comparison for Mozambique at the north betwee original resolution and 100m maps

Screenshot 2024-06-03 at 4 22 14 PM

Testing files 
[moz-geojson.tgz](https://github.com/user-attachments/files/15539419/moz-geojson.tgz). I used https://mapshaper.org/ to load them in separate layers.

matiasgarciaisaia (approved):
LGTM!

As a note for deployment, we'll probably need to manually fix the input file with the duplicate keys when running this in production.

